PR #4, opened 2021-09-27
Return maas_instance network interface information as output

mlipscombe:
This PR adds code to return the `network_interfaces` field in the schema as output from a `maas_instance`, and configures default connection information so that the file and remote-exec provisioners can connect without the need for a `connection` block.


skatsaounis:
Closed due to inactivity

PR #7, opened 2021-12-13
Add datasources for machine and physical nic

ianmcmahon:
Create datasources for existing machines and physical nics


maas/provider.go
```diff
@@ -53,6 +53,8 @@ func Provider() *schema.Provider {
 			"maas_fabric": dataSourceMaasFabric(),
 			"maas_vlan":   dataSourceMaasVlan(),
 			"maas_subnet": dataSourceMaasSubnet(),
+            "maas_machine": dataSourceMaasMachine(),
```
cpg1111, line 56: seems you need to run `go fmt`

cpg1111 (changes requested):
One small issue with missing format, otherwise, looks good.

skatsaounis:
Hi @ianmcmahon

I know it has been a while since you opened this PR but we are interested in adding these data sources to the provider. Would you still like to contribute to this PR or should we proceed with our implementation? If yes, feel free to reach out for help if needed :)

troyanov (approved):

maas/data_source_maas_machine.go
```diff
@@ -0,0 +1,82 @@
...
+func dataSourceMaasMachine() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceMachineRead,
+
+		Schema: map[string]*schema.Schema{
+			"power_type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"power_parameters": {
```
skatsaounis, line 20: This one will not work for machines with IPMI `power_parameters` that include `workaround_flags`. A similar approach to the `resource_maas_machine` should be followed here.

skatsaounis (changes requested):

maas/data_source_maas_machine.go
troyanov: Good catch!

ironashram:
@troyanov for my use case (and i suppose for many more) it would be needed to fetch the maas_machine datasource to with pxe_mac_address as input, since the hostname which is currently used here is not predictable in case the machine has been discovered automatically

My fork already has this, it may not be the cleanest but it works quite well

skatsaounis:
@ironashram I incorporated most of what you have in your fork, since it makes a lot of sense. At last, we are ready to run our e2e tests and get some approval here :tada:

skatsaounis (approved):

cpg1111 (approved):
LGTM

PR #20, opened 2022-08-24
Introduce GitHub workflows

trentrosenbaum:
This is a small PR to introduce the test and release GitHub action workflows.  These have been raised from the Terraform provider scaffolding repo.  The aim is to be a starting point that can be expanded.

The `release.yml` representing the release workflow will require additional environment variables to be setup in the repo and this can be expanded later.

The `test.yml` representing the test workflow will allow us to introduce the acceptance tests.  The idea is to have a basic acceptance test that covers the provider.go and the error handling around the environment variables.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,79 @@
+# Terraform Provider testing workflow.
+name: Tests
+
+# This GitHub action runs your tests for each pull request and push.
+# Optionally, you can turn it on using a schedule for regular testing.
+on:
+  pull_request:
+    paths-ignore:
+      - 'README.md'
+      - 'docs'
+      - 'examples'
+  push:
+    paths-ignore:
+      - 'README.md'
+      - 'docs'
+      - 'examples'
+
+# Testing only needs permissions to read the repository contents.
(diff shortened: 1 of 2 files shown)


pvegh-oc:
Target is master.

pvegh-oc (approved):

.github/workflows/test.yml
```diff
@@ -0,0 +1,79 @@
...
+          go-version-file: 'go.mod'
+          cache: true
+      - uses: hashicorp/setup-terraform@v2
+        with:
+          terraform_version: ${{ matrix.terraform }}
+          terraform_wrapper: false
+      - run: go mod download
+      - env:
+          TF_ACC: "1"
+        run: go test -v -cover ./internal/provider/
```
greg-oc, line 78:
```suggestion
        run: go test -v -cover ./maas/
```
I think this would match the structure of their repo, although it may be better to restructure the repo to fit the "/internal/provider/" template rather than changing this line

greg-oc (changes requested):

greg-oc (approved):

PR #29, opened 2022-09-12
Changes for Terraform Registry release

greg-oc:
### NEW:

* Add release process guide 
* Add GitHub Actions workflow for releasing
* Add Dependabot configuration

### IMPROVEMENTS:

* Update documentation and documentation structure to use terraform-plugin-docs tool
* Refactor API client references to use the "maas" repository
* Update dependencies
* Update Go version to 1.19


examples/data-sources/maas_fabric/data-source.tf
```diff
@@ -0,0 +1,3 @@
+data "maas_fabric" "default" {
+  name = "maas"
+}
```
cpg1111, line 3: also fix the line-end errors here

go.mod
```diff
@@ -1,12 +1,87 @@
 module terraform-provider-maas
 
-go 1.16
+go 1.19
```
cpg1111, line 3: is 1.19 actually required or would 1.18 suffice? The reason I ask is we tend to use the default deb installation of Go to build, and that happens to be 1.18 at the moment. We can work around that, but matching the versions would be easier.

cpg1111 (changes requested):
Handful of minor formatting issues, and a question regarding the required Go version.

cpg1111 (approved):
LGTM

PR #87, opened 2023-09-19
all: update Go module to Go 1.20

troyanov:
In order to use recent version of Terraform plugin SDK [0] we need to bump our minimal Go version to 1.20 because of this PR [1] being merged.

Also each major Go release is supported until there are two newer major releases [2].

[0]: github.com/hashicorp/terraform-plugin-sdk/v2
[1]: https://github.com/maas/terraform-provider-maas/pull/83
[2]: https://go.dev/doc/devel/release#policy


SK1Y101:
jenkins: !test

SK1Y101 (dismissed):

PR #96, opened 2023-10-04
Fix parsing of power parameters

skatsaounis:
This PR allows the handling of machines with power parameters that contain nested parameters, like IPMI `workaround_flags`.

Unfortunately, Terraform sdk cannot handle situations when a map contains elements of different types. Thus, with the existing implementation, we can handle only machines with power parameters which are `map[string]string`. As a [usual design workaround](https://github.com/hashicorp/terraform/issues/22346#issuecomment-519741511), the `power_parameters` is turned to a `schema.TypeString` which includes a serialized JSON.

With the above change in the field, we are able to manage (import, create, update) machines of type IPMI.


skatsaounis:
Consumes and depends on: https://github.com/maas/gomaasclient/pull/16

troyanov (approved):

PR #101, opened 2023-10-10
Add support for TLS configuration options

skatsaounis:
This PR adds support for TLS configuration options provided by the user. Being more specific, it introduces:

- The option to provide a path with a CA Cert to be used to communicate with MAAS
- The option to skip TLS verification, which by default is set to false


troyanov (approved):

pescobar:
this feature would be quite convenient for us

troyanov (approved):
LGTM

PR #108, opened 2023-10-16
all: drop support of Terraform EOL versions

troyanov:
1.4 is also EOL, however it is still widely used.

https://support.hashicorp.com/hc/en-us/articles/360021185113


skatsaounis (approved):
LGTM

PR #109, opened 2023-10-16
ci: special labels to trigger integration tests

troyanov:
Introduce label based workflow.
MAAS maintainers have to apply special label in order to trigger integration tests using real MAAS installation.

Labels are automatically removed when new commits are pushed to the PR branch.

- `Review: Core +1` label triggers execution of end-to-end tests


SK1Y101 (dismissed):
+1 from me

SK1Y101 (approved):
Still a +1 from me

Issue #98, opened 2023-10-07
feature req: add "maas_device" resource for creating unmanaged machines 


ednxzu:
It is currently not possible to manage resource of type : device with the provider.
Would it be possible to add this resource type into the provider ? 

If I missed something and it is possible to create such resources, please disregard this issue.

Thanks !

PR #119, opened 2023-11-16
Add support for resource and data source `maas_device`

skatsaounis:
This PR includes:

- A resource `maas_device` for managing devices
- A data source `maas_device` for reading existing devices
- Basic acceptance tests

Resolves: #98


maas/data_source_maas_device.go
```diff
@@ -0,0 +1,147 @@
...
+	return device, nil
+}
+
+func findDevice(client *client.Client, identifier string) (*entity.Device, error) {
+	devices, err := client.Devices.Get()
+	if err != nil {
+		return nil, err
+	}
+	for _, d := range devices {
+		if fmt.Sprintf("%v", d.SystemID) == identifier || d.Hostname == identifier {
```
troyanov, line 142: Why do we need `fmt.Sprintf` here, isn't `d.SystemID` of type `string`?

troyanov (approved):
LGTM with comment

maas/data_source_maas_device.go
skatsaounis: No reason at all. It was copied from another function without being careful. Fixed :)

troyanov (approved):

Issue #120, opened 2023-11-20
VLAN field of Physical interface should be the database id instead of name


skatsaounis:
While the `VLAN` field is tied to `schema.TypeString`, this is not correct and even the codebase of provider is handling it as a database id rather than the VLAN name of the interface.

The following example is trying to set VLAN to untagged, inspired by the field description that mentions default value of VLAN set to `untagged`.

```bash
terraform apply -auto-approve

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # maas_network_interface_physical.virsh_vm1_nic1 will be created
  + resource "maas_network_interface_physical" "virsh_vm1_nic1" {
      + id          = (known after apply)
      + mac_address = "52:54:00:89:f5:3e"
      + machine     = "eq6aqs"
      + mtu         = (known after apply)
      + name        = (known after apply)
      + tags        = [
          + "nic1-tag1",
          + "nic1-tag2",
          + "nic1-tag3",
        ]
      + vlan        = "untagged"
    }

Plan: 1 to add, 0 to change, 0 to destroy.
maas_network_interface_physical.virsh_vm1_nic1: Creating...
╷
│ Error: ServerError: 400 Bad Request ({"vlan": ["Select a valid choice. That choice is not one of the available choices."]})
│ 
│   with maas_network_interface_physical.virsh_vm1_nic1,
│   on network.tf line 1, in resource "maas_network_interface_physical" "virsh_vm1_nic1":
│    1: resource "maas_network_interface_physical" "virsh_vm1_nic1" {
```

Issue #116, opened 2023-11-15
Error -- json: cannot unmarshal number into Go struct field BlockDevice.physicalblockdevice_set.used_size of type int


ednxzu:
Basically what the title says, when provisioning a `maas_machine` resource, terraform throws an error related to the disk size 

provider version: `1.3.0` or `1.3.1`, both have the same behaviour

```
resource "maas_machine" "os_control_01" {
  hostname         = "os-control-01"
  domain           = "<redacted>"
  pool             = "openstack"
  power_parameters = {}
  power_type       = "manual"
  pxe_mac_address  = "<redacted>"
  zone             = "openstack"
}
```

throws 

```
│ Error: json: cannot unmarshal number 128032178176 into Go struct field BlockDevice.physicalblockdevice_set.used_size of type int
```

I'm not sure what additional info you might want, just tell me and I'll paste it here !

Cheers !

troyanov:
Hi @ednxzu 

May I ask you for the output of `uname -m` on the machine where you run Terraform provider?

ednxzu:
> Hi @ednxzu
> 
> May I ask you for the output of `uname -m` on the machine where you run Terraform provider?

Hey sorry I haven't got back to you, seems that you figured it out, thanks !

troyanov:
> Hey sorry I haven't got back to you, seems that you figured it out, thanks !

No worries. Hopefully the issue is fixed now.
It will take us some time to release a new version, but if you can try with the latest master, that would be great!

troyanov:
Ooops, this is still valid, until we update go.mod to point to the latest gomaasclient that includes https://github.com/maas/gomaasclient/pull/21 fix

ednxzu:
Hey, no worries ! Tho I noticed that this should be happening on 32 bits systems ? I ran into this error on my x86_64 workstation. 

For more context, With opentofu it throws this error, but terraform (the real one) doesn't. So I just reverted to using terraform for this. 

Maybe this is also a bug in tofu ?

skatsaounis:
Hi @ednxzu, with the merge of #121 we expect that your issue is no longer present in the provider binary created by the latest master branch. Since we are not ready yet to issue a new release, if you would like to try from now, you can build the provider locally and confirm that the error is gone. I am closing the issue for now and if we are still missing any bit we will reopen accordingly. Thank you for your engagement so far :beers:

PR #121, opened 2023-11-20
Consume latest gomaasclient and fix drifts

skatsaounis:

- Disable acceptance tests since the intention is to run internally by Jenkins job
- Run unit tests in a GitHub runner job
- Change VLAN from string to integer and properly document that it refers to the database ID of the VLAN
- Bump `gomaasclient` to the latest master version
- Fix `int` types that changed to `int64` in the gomaasclient
- Change the order of fields to be in alphabetical order wherever they are present


skatsaounis:
Addresses: #116, #120

troyanov (dismissed):

troyanov (approved):
Nice! Thanks for explaining `int64(d.GetField().int())`

Maybe we could drop couple links into comments here, so others can learn it as well.

skatsaounis:
Sure thing.

First of all, this PR is introducing the usage of `UseJSONNumber` in every resource that contains fields that may include big numbers. The description of the [PR](https://github.com/hashicorp/terraform-plugin-sdk/pull/662) that introduced the flag to the SDK is well describing its purpose. Big providers like GCP are using the flag as well.

In addition, Terraform SDK does not have a separate field type for int64. Instead, it is using `pluginsdk.TypeInt` for any number. In situations where Go clients are using `int64` struct fields which are reflected to `pluginsdk.TypeInt` Terraform provider fields the assignments between them are the following:

- provider SDK to Go client struct field: `machine.Memory = int64(d.Get("memory").(int))`
- Go client struct field to provider SDK: `d["memory"] = machine.Memory`

where `` Memory int64 `json:"memory,omitempty"` ``, `"memory": {Type: schema.TypeInt,}`

Many examples of those assignments can also be found on big providers like AWS, Azure, GCP.

Issue #135, opened 2024-01-05
Incorrect version string in makefile


SK1Y101:
the current makefile cites itself as very 1.0.1.
We've already released version 2.0.

The release automation should probably update this too.

skatsaounis:
Thanks for opening this one Jack but hopefully it is not a bug. This version is not related with the actual version of the provider. Instead, it is a statically set version which is needed when running provider from a custom binary of a specific branch/commit. For someone to be able to use a binary from a commit all is needed is a proper import of the file created by the Makefile:

```tf
terraform {
  required_providers {
    maas = {
      source  = "registry.terraform.io/maas/maas"
      version = "=1.0.1"
    }
  }
}
```

With the above snippet, user is asking Terraform to use maas provider, and based on the `source` value, Terraform will search this provider locally in the following path: `~/.terraform.d/plugins/registry.terraform.io/maas/maas/1.0.1/linux_amd64/terraform-provider-maas`, in case of linux amd64. Makefile is making sure that the binary will be there: https://github.com/maas/terraform-provider-maas/blob/master/Makefile#L32. As a convention, a version is needed, so an arbitrary is chosen.

PR #136, opened 2024-01-05
Update version string in makefile

SK1Y101:
We've released Version 2.0, but this string is still set to 1.0.1. That should be changed


SK1Y101:
fixes #135

Issue #137, opened 2024-01-06
Deleting a VM Host of a dynamic machine occurs in error


skatsaounis:
MAAS can accept credentials of an existing KVM or LXD VM host and when those are provided, the VM host is added successfully. However, there is a big difference between this type of VM hosts and those that are created from deployed MAAS machines with the relevant option set.

To accommodate their details, MAAS is creating a dynamic type machine, which is not an actual machine and there is logic that is deleting it when VM host is deleted. This information is not exposed to the API, so `gomaasclient` cannot fetch it and as such, the Terraform provider cannot be aware of it. However, Terraform provider could ask MAAS after the deletion of the VM host if its linked machine is still present or has been deleted.

The existing implementation of Terraform provider assumes that VM hosts of dynamic machines do not include their system id in their details. By making this wrong assumption, it tries to release the machine of any VM host that was deleted and included its system id in their details. However, since this check is wrong and VM hosts of dynamic machines are also including their system ids in their details, the release occurs in error to Terraform provider delete action. Unfortunately, MAAS has already deleted the dynamic machine.

skatsaounis:
Implementing this will also unblock the solution for: https://github.com/maas/terraform-provider-maas/issues/137

PR #138, opened 2024-01-06
fix: deleting VM hosts of dynamic machines do not try release

skatsaounis:
Addresses #137 and also improves the update by removing the unnecessary VM host get call

### DIFF
--- a/maas/resource_maas_vm_host.go
+++ b/maas/resource_maas_vm_host.go
@@ -3,12 +3,14 @@ package maas
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"strconv"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/juju/gomaasapi/v2"
 	"github.com/maas/gomaasclient/client"
 	"github.com/maas/gomaasclient/entity"
 )
@@ -225,13 +227,9 @@ func resourceVMHostUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 	if err != nil {
 		return diag.FromErr(err)
 	}
-	vmHost, err := client.VMHost.Get(id)
-	if err != nil {
-		return diag.FromErr(err)
-	}
 
 	// Update VM host options
-	_, err = client.VMHost.Update(vmHost.ID, getVMHostParams(d))
+	_, err = client.VMHost.Update(id, getVMHostParams(d))
 	if err != nil {
 		return diag.FromErr(err)
 	}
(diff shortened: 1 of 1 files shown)


maas/resource_maas_vm_host.go
```diff
@@ -225,13 +227,9 @@ func resourceVMHostUpdate(ctx context.Context, d *schema.ResourceData, meta inte
...
 		return diag.FromErr(err)
 	}
-	vmHost, err := client.VMHost.Get(id)
-	if err != nil {
-		return diag.FromErr(err)
-	}
 
 	// Update VM host options
-	_, err = client.VMHost.Update(vmHost.ID, getVMHostParams(d))
+	_, err = client.VMHost.Update(id, getVMHostParams(d))
```
troyanov, line 232: My understanding that this change is just an optimisation to remove extra API call under `client.VMHost.Get`?

maas/resource_maas_vm_host.go
```diff
@@ -256,20 +254,27 @@ func resourceVMHostDelete(ctx context.Context, d *schema.ResourceData, meta inte
...
-			return diag.FromErr(err)
-		}
-		// Wait machine to be released
-		_, err = waitForMachineStatus(ctx, client, vmHost.Host.SystemID, []string{"Releasing"}, []string{"Ready"})
-		if err != nil {
-			return diag.FromErr(err)
+	// Check if VM host was linked to a dynamic machine and if yes, return
+	// Dynamic machines are deleted by MAAS when their VM hosts are deleted.
+	// This information is not directly available from the API.
+	_, err = client.Machine.Get(vmHost.Host.SystemID)
```
troyanov, line 260: I don't understand what you are trying to achieve here.

A VM host is a machine designed to run virtual machines.
`vmHost.Host.SystemID` will contain `systemID` of a physical machine that was deployed with "Register as MAAS KVM host" option, right?

My understanding is that we want to delete all _linked_ VMs that are linked to a VM host. Or am I wrong?

troyanov (commented):
@skatsaounis not sure I understand the change. Can you please clarify a bit? Thanks!

maas/resource_maas_vm_host.go
skatsaounis: This is true, yes

maas/resource_maas_vm_host.go
skatsaounis: VM Host can also be an existing VM Host aka a machine not created by MAAS but is running LXD or Virsh and we register its credentials so MAAS can register it as a VM Host. To achieve that, MAAS is using the concept of dynamic machine. For example, try to add your laptop as a VM Host to your lab MAAS. Obviously the laptop is not a machine owned by MAAS, but you can see it being displayed as a maas machine on machines tab. This is a dynamic machine and when VM Host is deleted, that abstraction of a machine is also deleted by MAAS with specific logic. In such cases, Terraform will get an error because it will try to ask MAAS to release your laptop, since `systemID` is pointing to something that does not exist. #137 has some additional info.

skatsaounis: This is not related with linked VMs of a VM Host at all. Instead it is taking care of the machine that acts as a VM host which can be either MAAS owned or something external represented as dynamic machine

Issue #132, opened 2023-12-14
Default distro_series "focal" causing deployment failure in cases where focal image does not exist


vishvikkrishnan:
The maas_vm_host resource type causes an error during `terraform apply` when trying to deploy a machine in MAAS without the "focal" image. Consequently, it is currently not possible to deploy a pre-registered MAAS machine with any image other than "focal" using the `machine` param:

```hcl
Terraform will perform the following actions:

  # maas_vm_host.infra_2_lxd_host will be created
  + resource "maas_vm_host" "infra_2_lxd_host" {
      + cpu_over_commit_ratio         = (known after apply)
      + default_macvlan_mode          = (known after apply)
      + id                            = (known after apply)
      + machine                       = "cos-2"
      + memory_over_commit_ratio      = (known after apply)
      + name                          = "infra-2-lxd-host"
      + pool                          = (known after apply)
      + resources_cores_total         = (known after apply)
      + resources_local_storage_total = (known after apply)
      + resources_memory_total        = (known after apply)
      + tags                          = (known after apply)
      + type                          = "lxd"
      + zone                          = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

maas_vm_host.infra_2_lxd_host: Creating...
╷
│ Error: ServerError: 400 Bad Request ({"distro_series": ["'focal' is not a valid distro_series.  It should be one of: '', 'ubuntu/jammy'."]})
│ 
│   with maas_vm_host.infra_2_lxd_host,
│   on main.tf line 16, in resource "maas_vm_host" "infra_2_lxd_host":
│   16: resource "maas_vm_host" "infra_2_lxd_host" {
╵
```

MAAS **only** has the `jammy` image by default. Configuration used:
```hcl
resource "maas_vm_host" "infra_2_lxd_host" {
  type = "lxd"
  name = "infra-2-lxd-host"
  machine = "cos-2" #machine is pre-registered in MAAS
}
```
This is because the default "distro_series" is hardcoded to "focal" in the [source file](https://github.com/maas/terraform-provider-maas/blob/76a4771b604fdc7f2e34999776c7f3b0a68a6dac/maas/resource_maas_vm_host.go#L308).

[...]

PR #142, opened 2024-01-15
fix: use MAAS defaults for VM host provisioning

skatsaounis:
Instead of using statically set `focal`, use MAAS defaults.

### DIFF
--- a/maas/resource_maas_vm_host.go
+++ b/maas/resource_maas_vm_host.go
@@ -2,6 +2,7 @@ package maas
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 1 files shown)


skatsaounis:
Fixes: #132

troyanov (approved):
LGTM
@skatsaounis may I ask you to address `gomaasclient` version bump in a separate PR please?

maas/resource_maas_vm_host.go
```diff
@@ -303,9 +304,29 @@ func deployMachineAsVMHost(ctx context.Context, client *client.Client, machineId
...
+		fmt.Println(err)
+	}
+	var defaultSeries string
+	defaultSeriesBytes, err := client.MAASServer.Get("default_distro_series")
+	if err != nil {
+		fmt.Println(err)
+	}
+	err = json.Unmarshal(defaultSeriesBytes, &defaultSeries)
+	if err != nil {
+		fmt.Println(err)
```
troyanov, line 324: Maybe instead printing an error  we should return `nil, err`?
Or will it eventually fail because it'll try to use emtpy OS and distro?
 ```go
DistroSeries:   fmt.Sprintf("%s/%s", defaultOsystem, defaultSeries)
```

skatsaounis: you are right. The fmts were leftovers from a playground code that I maintain. I replaced them with returns. Good catch :+1:

troyanov (approved):

PR #164, opened 2024-03-18
feat: add configurable create timeout for VM hosts

SK1Y101:
VM Host operations can take longer than the alloted 20 minutes, this should introduce a user configurable `timeouts` block to increase that time if needed.

### DIFF
--- a/maas/resource_maas_instance.go
+++ b/maas/resource_maas_instance.go
@@ -205,7 +205,7 @@ func resourceMaasInstance() *schema.Resource {
 			},
 		},
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(40 * time.Minute),
+			Create: schema.DefaultTimeout(20 * time.Minute),
 		},
 	}
 }
--- a/maas/resource_maas_instance.go
+++ b/maas/resource_maas_instance.go
@@ -205,7 +205,7 @@ func resourceMaasInstance() *schema.Resource {
 			},
 		},
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(20 * time.Minute),
+			Create: schema.DefaultTimeout(40 * time.Minute),
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


maas/resource_maas_instance.go
```diff
@@ -205,7 +205,7 @@ func resourceMaasInstance() *schema.Resource {
 			},
 		},
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(20 * time.Minute),
+			Create: schema.DefaultTimeout(40 * time.Minute),
```
skatsaounis, line 208: I would suggest to revert this change. Being more specific, what this line does is giving the option to our users to explicitly set a timeout when creating `maas_instance` resources. It is also setting the default `create` timeout in case users do not want to set something custom.

If `Timeouts` field is not declared in a resource, Terraform will use the default timeouts for each operations: https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/resource_data.go#L412-L445. As you can see, SDK default choice is also 20 minutes, which I find good.

My suggestion is to bump the timeouts on the resource declarations in the tests that are failing. This can be done with something like this:

```tf
resource "maas_instance" "tf_test_vm_instance" {
  allocate_params {
    hostname = maas_vm_host_machine.tf_test_vm.hostname
  }
  deploy_params {
    distro_series = var.distro_series
  }
  timeouts {
    create = "40m"
  }
}
```

But what we really miss is the ability to declare custom create timeouts on `maas_vm_host` resources. This is extremely useful when we need MAAS to deploy a machine and use it as a VM Host. So I would love to add this with this PR. By the way, this is the problematic resource in our internal tests.

I am more than happy to assist you with the review :slightly_smiling_face:

skatsaounis (changes requested):

maas/resource_maas_instance.go
skatsaounis: Here is a link from the SDK documentation if you need to explore more the timeouts: https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts

SK1Y101: I think I've added to the correct place instead?

skatsaounis: Your change looks good. Let's wait for a green CI run before merging it :)

skatsaounis (approved):
LGTM +1

skatsaounis:
@SK1Y101 could you please update the PR description and merge it? Our e2e tests are happy and green :slightly_smiling_face:

Issue #163, opened 2024-03-08
Bug: Importing and renaming physical interface tends to remove machine instead


vishvikkrishnan:
When trying to import an existing physical interface of machines into the plan, so it can be modified using the `maas_network_interface_physical` resource, the plan instead warns that the machines would be replaced since the `machine` key doesn't resolve the values of `system_id`, `hostname`, or `fqdn` before applying the plan.

TF config to import the nic:
```hcl
import {
  to = maas_network_interface_physical.configure_nic
  id = "node01:enx8cae4cdf507c"
}

resource "maas_network_interface_physical" "configure_nic" {  
  machine = var.nic.machine
  mac_address = var.nic.mac_address
  name = "usb0"
}
```

The following error shows on trying to generate the plan:
```shell
 # maas_network_interface_physical.configure_nic["node01"] must be replaced
  # (imported from "node01:enx8cae4cdf507c")
  # Warning: this will destroy the imported resource
-/+ resource "maas_network_interface_physical" "configure_nic" {
      ~ id          = "131" -> (known after apply)
        mac_address = "8c:ae:4c:df:50:7c"
      ~ machine     = "b8qefh" -> "node01" # forces replacement
      ~ mtu         = 1500 -> (known after apply)
      ~ name        = "enx8cae4cdf507c" -> "usb0"
      ~ tags        = [] -> (known after apply)
      - vlan        = 58 -> null
    }
```
In the above error, the provider does not resolve the `hostname` with the `system_id` of the node. This is not the expected behavior since the provider documentation mentions that [any of the three machine identifiers](https://registry.terraform.io/providers/maas/maas/latest/docs/resources/network_interface_physical#:~:text=machine%20(String)%20The%20identifier%20(system%20ID%2C%20hostname%2C%20or%20FQDN)%20of%20the%20machine%20with%20the%20physical%20network%20interface.) can be provided.

Also, the import block only accepts the name of the interface (enx8cae4cdf507c) instead of also accepting the MAC address. Since the main purpose of importing the network interface is to rename it (from randomly generated names), it would be nice to be able to import the interface using the MAC address as well.

baycarbone:
fyi, this error seems to be affecting only the maas_network_interface_physical resource; with maas_network_interface_link, referencing a machine using its hostname works fine.

skatsaounis:
[...]

PR #169, opened 2024-04-11
fix: properly import physical interfaces

skatsaounis:

- Change the delimiter of importing physical interfaces from `:` to `/`
- Allow import by MAC Address as advertised
- Do not break idempotency when importing existing resources
- Add acceptance tests for physical interfaces

Fixes: #163

### DIFF
--- a/maas/resource_maas_network_interface_physical.go
+++ b/maas/resource_maas_network_interface_physical.go
@@ -21,9 +21,9 @@ func resourceMaasNetworkInterfacePhysical() *schema.Resource {
 		DeleteContext: resourceNetworkInterfacePhysicalDelete,
 		Importer: &schema.ResourceImporter{
 			StateContext: func(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-				idParts := strings.Split(d.Id(), ":")
+				idParts := strings.Split(d.Id(), "/")
 				if len(idParts) != 2 || idParts[0] == "" || idParts[1] == "" {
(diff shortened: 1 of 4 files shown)


maas/resource_maas_network_interface_physical.go
```diff
@@ -34,18 +34,8 @@ func resourceMaasNetworkInterfacePhysical() *schema.Resource {
...
-					"mtu":         n.EffectiveMTU,
-					"name":        n.Name,
-					"tags":        n.Tags,
-					"vlan":        n.VLAN.ID,
-				}
-				if err := setTerraformState(d, tfState); err != nil {
-					return nil, err
-				}
+				d.Set("machine", idParts[0])
+				d.SetId(fmt.Sprintf("%v", n.ID))
```
troyanov, line 38: Considering that [ID](https://github.com/maas/gomaasclient/blob/78ffc951871382b09ea39a96eed24964df038b46/entity/network_interface.go#L25) is an `int` we can use https://pkg.go.dev/strconv#Itoa

More on this topic https://dev.to/mavensingh/strconvitoa-vs-strconvformatint-vs-fmtsprintf-which-is-good-to-convert-int-to-string-in-golang-3iol

troyanov (approved):
LGTM with a small nitpick

maas/resource_maas_network_interface_physical.go
skatsaounis: Thank you for the suggestion. I also applied this to all other usages of Sprintf in this file

troyanov (approved):

PR #178, opened 2024-05-20
fix: manage existing physical interfaces from commissioning

### LINKED ISSUE #165: Bug: Creating new physical network interface with a pre-defined name fails to rename the interface after apply
The `terraform plan` lists that the name would be applied to the created physical network interface as follows:

```hcl
# maas_network_interface_physical.configure_nic["node01"] will be created
  + resource "maas_network_interface_physical" "configure_nic" {
      + id          = (known after apply)
      + mac_address = "8c:ae:4c:df:50:7c"
      + machine     = (known after apply)
      + mtu         = (known after apply)
      + name        = "usb0"
      + tags        = (known after apply)
    }
```

However, after the `terraform apply`, the state indicates shows the resource with the original interface name as follows:

```hcl
# maas_network_interface_physical.configure_nic["node01"]:
resource "maas_network_interface_physical" "configure_nic" {
    id          = "176"
    mac_address = "8c:ae:4c:df:50:7c"
    machine     = "73ywba"
    mtu         = 1500
    name        = "enx8cae4cdf507c"
    tags        = []
    vlan        = 89
}
```

Additionally, trying to reapply the same plan, does update the name through an in-place update operation.

```hcl
  # maas_network_interface_physical.configure_nic["node01"] will be updated in-place
  ~ resource "maas_network_interface_physical" "configure_nic" {
        id          = "176"
      ~ name        = "enx8cae4cdf507c" -> "usb0"
        tags        = []
        # (4 unchanged attributes hidden)
    }
```
The state is as expected after re-applying the same plan:

```hcl
# maas_network_interface_physical.configure_nic["node01"]:
resource "maas_network_interface_physical" "configure_nic" {
    id          = "176"
    mac_address = "8c:ae:4c:df:50:7c"
    machine     = "73ywba"
    mtu         = 1500
    name        = "usb0"
    tags        = []
    vlan        = 17
}
```

The name should ideally update during the first apply as well, and there should be no need to apply the same plan again (otherwise, there's no idempotency.)

skatsaounis:

- add `network_interfaces` calculated field on `maas_machine`. The field includes a set of MAC addresses of discovered interfaces during commissioning.
- change logic of `maas_network_interface_physical` when another interface is found with the same MAC address  during creation. From now on, an update operation is tried instead of no op.
- improve the logic of create, update actions. Since MAAS API is returning the created/updated object there is no need to call read action before returning. By using this improvement, we avoid running the expensive operations of `client.Machines.Get` and `client.NetworkInterface.Get`.
- make `vlan` field of `maas_network_interface_physical` "computable". Interfaces detected during commissioning already belong to a VLAN and this should be detected during the Terraform resource creation. Otherwise, idempotency is broken.

This PR fixes #165

### DIFF
--- a/docs/resources/machine.md
+++ b/docs/resources/machine.md
@@ -45,6 +45,7 @@ resource "maas_machine" "virsh_vm1" {
 ### Read-Only
 
 - `id` (String) The ID of this resource.
+- `network_interfaces` (Set of String) A set of MAC addresses of network interfaces attached to the machine.
 
 <a id="nestedblock--timeouts"></a>
 ### Nested Schema for `timeouts`
--- a/maas/resource_maas_machine.go
+++ b/maas/resource_maas_machine.go
@@ -85,6 +85,14 @@ func resourceMaasMachine() *schema.Resource {
 				Computed:    true,
 				Description: "The minimum kernel version allowed to run on this machine. Only used when deploying Ubuntu. This is computed if it's not set.",
 			},
+			"network_interfaces": {
+				Type:        schema.TypeSet,
+				Computed:    true,
+				Description: "A set of MAC addresses of network interfaces attached to the machine.",
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
 			"pool": {
 				Type:        schema.TypeString,
 				Optional:    true,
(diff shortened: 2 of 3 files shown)


docs/resources/machine.md
```diff
@@ -53,6 +54,15 @@ Optional:
 
 - `create` (String)
 
+
+<a id="nestedatt--network_interfaces"></a>
+### Nested Schema for `network_interfaces`
+
+Read-Only:
+
+- `id` (Number)
```
troyanov, line 63: It is considered to be a good practice not to return database internal identifiers. I cannot think about use-case for ID, maybe it is better not to return it at all? Maybe in future we'll move to GUID for all the resources in MAAS

troyanov (approved):
LGTM with a comment about ID

PR #208, opened 2024-07-03
ci: fix goreleaser

skatsaounis:
Based on goreleaser [deprecations](https://goreleaser.com/deprecations/#__tabbed_17_1), `--rm-dist` is replaced by `--clean`.


skatsaounis:
Integration tests can be skipped since the change is irrelevant with the provider itself.

PR #210, opened 2024-07-03
chore: update Go module to Go 1.21

skatsaounis:
Each major Go release is supported until there are two newer major releases [0].

[0]: https://go.dev/doc/devel/release#policy

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ References:
 ## Requirements
 
 - [Terraform](https://www.terraform.io/downloads.html) >= 1.4.x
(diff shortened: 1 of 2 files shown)


troyanov:
@skatsaounis this might be a breaking change for some people that still build provider with Go 1.20

It would be better to postpone minimal Go version bump until we will start using new language features.
For now we can just bump version of `actions/setup-go` (to use latest & greatest Go compiler to build artefacts)

What do yo think?

troyanov:
https://github.com/canonical/terraform-provider-maas/pull/211/files

Issue #206, opened 2024-07-02
Modifications to VLAN interfaces crash provider


sempervictus:
```hcl

  # maas_network_interface_vlan.metal-1["test-metal1-00"] will be updated in-place
  ~ resource "maas_network_interface_vlan" "metal-1" {
        id        = "16"
      ~ mtu       = 1500 -> 9000
        tags      = []
        # (5 unchanged attributes hidden)
    }

  # maas_network_interface_vlan.metal-1["test-metal1-01"] will be updated in-place
  ~ resource "maas_network_interface_vlan" "metal-1" {
        id        = "14"
      ~ mtu       = 1500 -> 9000
        tags      = []
        # (5 unchanged attributes hidden)
    }

  # maas_network_interface_vlan.metal-1["test-metal1-02"] will be updated in-place
  ~ resource "maas_network_interface_vlan" "metal-1" {
        id        = "15"
      ~ mtu       = 1500 -> 9000
        tags      = []
        # (5 unchanged attributes hidden)
    }

  # maas_network_interface_vlan.metal-1["test-metal1-03"] will be updated in-place
  ~ resource "maas_network_interface_vlan" "metal-1" {
        id        = "17"
      ~ mtu       = 1500 -> 9000
        tags      = []
        # (5 unchanged attributes hidden)
    }

...
╷
│ Warning: Deprecated Resource
│ 
│   with openstack_compute_floatingip_associate_v2.test-chef00_priv,
│   on chef00.tf line 46, in resource "openstack_compute_floatingip_associate_v2" "test-chef00_priv":
│   46: resource "openstack_compute_floatingip_associate_v2" "test-chef00_priv" {
│ 
│ use openstack_networking_floatingip_associate_v2 resource instead
│ 
│ (and 7 more similar warnings elsewhere)
╵

Do you want to perform these actions?
  OpenTofu will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

maas_network_interface_vlan.metal-1["test-metal1-03"]: Modifying... [id=17]
maas_network_interface_vlan.metal-1["test-metal1-01"]: Modifying... [id=14]
maas_network_interface_vlan.metal-1["test-metal1-02"]: Modifying... [id=15]
maas_network_interface_vlan.metal-1["test-metal1-00"]: Modifying... [id=16]
╷
│ Error: Request cancelled
│ 
│   with maas_network_interface_vlan.metal-1["test-metal1-00"],
[...]

PR #227, opened 2024-07-29
fix: correctly parse vlan field of VLAN interfaces

skatsaounis:
- Parse `vlan` field as integer rather than string
- Add acceptance test to check the context update of all interface types

Resolves GH: #206

### DIFF
--- a/maas/resource_maas_network_interface_vlan.go
+++ b/maas/resource_maas_network_interface_vlan.go
@@ -87,7 +87,7 @@ func resourceNetworkInterfaceVlanCreate(ctx context.Context, d *schema.ResourceD
 	if err != nil {
 		return diag.FromErr(err)
 	}
-	vlan, err := getVlan(client, fabric.ID, fmt.Sprintf("%v", d.Get("vlan").(int)))
+	vlan, err := getVlan(client, fabric.ID, strconv.Itoa(d.Get("vlan").(int)))
 	if err != nil {
 		return diag.FromErr(err)
 	}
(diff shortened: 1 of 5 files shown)


SK1Y101 (approved):

maas/resource_maas_network_interface_vlan.go
```diff
@@ -167,7 +167,7 @@ func resourceNetworkInterfaceVlanUpdate(ctx context.Context, d *schema.ResourceD
 		return diag.FromErr(err)
 	}
 
-	vlan, err := getVlan(client, fabric.ID, d.Get("vlan").(string))
+	vlan, err := getVlan(client, fabric.ID, fmt.Sprintf("%v", d.Get("vlan").(int)))
```
troyanov, line 170: What is the benefit over previous approach here?
`fmt.Sprintf("%v", d.Get("vlan").(int))` vs `d.Get("vlan").(string)`

troyanov (dismissed):
LGTM with one question

maas/resource_maas_network_interface_vlan.go
troyanov: https://pkg.go.dev/strconv#Itoa

skatsaounis: This is actually the fix for: `panic: interface conversion: interface {} is int, not string`. Good catch on the less performant usage of `fmt.Sprintf`. Let me update both create and update functions to use Itoa

troyanov (approved):

Issue #128, opened 2023-12-12
Missing functionality for managing boot sources and boot resources (images)


vishvikkrishnan:
Need resource types to edit boot source, and manage its boot resources based on selection. Functionality to import images from a selected boot source is also missing.

It would be helpful to have a MAAS Terraform provider implementation for, at least, the following operations/commands:

[Editing a boot source](https://maas.io/docs/working-with-standard-images#heading--edit-a-boot-source)
`maas $PROFILE boot-source update $SOURCE_ID url=$URL keyring_filename=$KEYRING_FILE`

[Importing newly selected images for the boot source](https://maas.io/docs/working-with-standard-images#heading--import-newly-selected-images)
`maas $PROFILE boot-resources import`

PR #253, opened 2025-01-29
feat: add Boot source Data source

SK1Y101:
Implements the data-source side of #128 

A follow up PR should implement the CRUD actions

### DIFF
--- a/maas/data_source_maas_boot_source_test.go
+++ b/maas/data_source_maas_boot_source_test.go
@@ -15,7 +15,7 @@ func TestAccDataSourceMaasBootSource_basic(t *testing.T) {
 	checks := []resource.TestCheckFunc{
 		resource.TestCheckResourceAttr("data.maas_boot_source.test", "url", url),
 		resource.TestCheckResourceAttrSet("data.maas_boot_source.test", "created"),
-		// resource.TestCheckNoResourceAttr("data.maas_boot_source.test", "keyring_data"),
+		resource.TestCheckResourceAttr("data.maas_boot_source.test", "keyring_data", ""),
 		resource.TestCheckResourceAttr("data.maas_boot_source.test", "keyring_filename", keyring_path),
 		resource.TestCheckResourceAttrSet("data.maas_boot_source.test", "updated"),
 	}


maas/data_source_maas_boot_source_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+func TestAccDataSourceMaasBootSource_basic(t *testing.T) {
+
+	url := "https://images.maas.io/ephemeral-v3/candidate/"
+
+	checks := []resource.TestCheckFunc{
+		resource.TestCheckResourceAttr("data.maas_boot_source.test", "url", url),
+		resource.TestCheckResourceAttrSet("data.maas_boot_source.test", "created"),
+		resource.TestCheckResourceAttrSet("data.maas_boot_source.test", "updated"),
+		resource.TestCheckResourceAttrSet("data.maas_boot_source.test", "keyring_data"),
```
skatsaounis, line 19: This is the response that we get with snap installation and the defaults. This type of installation is also used by the system-tests that we perform. As such, this acceptance test will fail. based on that I would check that the keyring_data is not set and also the keyring_filename has an exact match with `/snap/maas/current/usr/share/keyrings/ubuntu-cloudimage-keyring.gpg`. In addition, line 13 should change to point to `stable` images

```
$ maas local boot-source read 2
Success.
Machine-readable output follows:
{
    "created": "2022-10-11T01:34:38.448",
    "updated": "2025-01-30T08:04:12.222",
    "url": "http://images.maas.io/ephemeral-v3/stable/",
    "keyring_filename": "/snap/maas/current/usr/share/keyrings/ubuntu-cloudimage-keyring.gpg",
    "keyring_data": "",
    "id": 2,
    "resource_uri": "/MAAS/api/2.0/boot-sources/2/"
}
```

maas/data_source_maas_boot_source_test.go
```diff
@@ -0,0 +1,43 @@
...
+				Config: testAccDataSourceMaasBootSource(url),
+				Check:  resource.ComposeTestCheckFunc(checks...),
+			},
+		},
+	})
+}
+
+func testAccDataSourceMaasBootSource(url string) string {
+	return fmt.Sprintf(`
+
```
skatsaounis, line 38: nit

```suggestion
```

maas/data_source_maas_boot_source.go
```diff
@@ -0,0 +1,68 @@
...
+				Description: "The filename on the keyring for the boot source.",
+			},
+			"updated": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The time of most recent update of the boot source.",
+			},
+			"url": {
+				Type:        schema.TypeString,
+				Required:    true,
```
skatsaounis, line 40: Considering that while MAAS allows you to have many boot sources, we assume in many places of its codebase that we can have only one, I was thinking that maybe URL can also be computed. In addition, with the data source we could return the only one boot source. We can just run `client.BootSources.Get()` and return the first item in the list, after checking that there is only one. An error should be thrown otherwise. Let's consider it.

skatsaounis (changes requested):
The PR is in a very good condition. However, I am adding 1-2 points to consider before starting testing it.

examples/data-sources/maas_boot_source/data-source.tf
```diff
@@ -0,0 +1,3 @@
+data "maas_boot_source" "default" {
```
skatsaounis, line 1: Since the URL is read-only

```suggestion
data "maas_boot_source" "default" {}
```

skatsaounis: And you also need to re-run the creation of the documentation

maas/data_source_maas_boot_source_test.go
```diff
@@ -0,0 +1,43 @@
...
+				Check:  resource.ComposeTestCheckFunc(checks...),
+			},
+		},
+	})
+}
+
+func testAccDataSourceMaasBootSource(url string) string {
+	return fmt.Sprintf(`
+data "maas_boot_source" "test" {
+	url = "%s"
```
skatsaounis, line 40: This functions should also change to not expect the URL as it is read-only

maas/resource_maas_boot_source.go
```diff
@@ -0,0 +1,30 @@
...
+		return nil, err
+	}
+	if len(bootsources) > 0 {
+		return &bootsources[0], nil
+	}
+	return nil, nil
+}
+
+func getBootSource(client *client.Client) (*entity.BootSource, error) {
+	bootsource, err := findBootSource(client)
```
skatsaounis, line 22: what if we simplify it a bit? Let's remove the second function findBootSource, return always the first bootsource found and also fail if the length is different than 1? The error is something like "expected a single boot source" or "cannot support more than one", or something similar?

skatsaounis (changes requested):
An extra set of suggestions. I think we are good to go after applying them

maas/data_source_maas_boot_source.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/canonical/gomaasclient/client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceMaasBootSource() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceMaasBootSourceRead,
+		Description: "Provides a resource to fetch MAAS boot sources.",
```
skatsaounis, line 15: I think after the change to fetch the one and only boot source, it makes sense to update the description like this. Note that this change will require re-generating the docs

```suggestion
		Description: "Provides a resource to fetch the MAAS boot source.",
```

skatsaounis (approved):
LGTM +1 🚀

skatsaounis (dismissed):

skatsaounis (approved):

PR #255, opened 2025-02-04
feat: add maas_boot_source resource

SK1Y101:
Adds the resource endpoints for #128 
- Add CRUD endpoints for boot_sources
- Add installation method field informing terraform how MAAS was installed ("snap" by default, or "deb")
- Add a wrapper around client.client (clientConfig) that includes a reference to the installation method
- Add testing for resource endpoints, which assumes the install_method is snap.


maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,126 @@
...
+	return resourceBootSourceRead(ctx, d, meta)
+}
+
+func resourceBootSourceDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
+	id, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	return diag.FromErr(client.BootSource.Delete(id))
```
skatsaounis, line 121: Based on the assumption that there is always one boot source that cannot be replaced, etc. the delete function should just revert the URL to point to the default URL that MAAS is setting, that is the URL of stable images

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,126 @@
...
+		ReadContext: dataSourceMaasBootSourceRead,
+		Description: "Provides a resource to fetch the MAAS boot source.",
+
+		Schema: map[string]*schema.Schema{
+			"created": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The creation time of the boot source.",
+			},
+			"keyring_data": {
```
skatsaounis, line 25: `keyring_data` and `keyring_filename` are mutually exclusive. Please take a look at `ConflictsWith` and how it can be used. This is an example from the MAAS provider code:

- https://github.com/canonical/terraform-provider-maas/blob/master/maas/resource_maas_vm_host.go#L108
- https://github.com/canonical/terraform-provider-maas/blob/master/maas/resource_maas_vm_host.go#L175

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,126 @@
...
+				Description: "The filename on the keyring for the boot source.",
+			},
+			"updated": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The time of most recent update of the boot source.",
+			},
+			"url": {
+				Type:        schema.TypeString,
+				Computed:    true,
```
skatsaounis, line 42: Since this is a field that we expect user to provide.

```suggestion
				Required:    true,
```

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,126 @@
...
+				Description: "The creation time of the boot source.",
+			},
+			"keyring_data": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The data on the keyring for the boot source.",
+			},
+			"keyring_filename": {
+				Type:        schema.TypeString,
+				Computed:    true,
```
skatsaounis, line 32:
```suggestion
				Optional:    true,
```

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,126 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"created": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The creation time of the boot source.",
+			},
+			"keyring_data": {
+				Type:        schema.TypeString,
+				Computed:    true,
```
skatsaounis, line 27:
```suggestion
				Optional:    true,
```

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,126 @@
...
 	"github.com/canonical/gomaasclient/client"
 	"github.com/canonical/gomaasclient/entity"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+func resourceMAASBootSource() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceMaasBootSourceRead,
+		Description: "Provides a resource to fetch the MAAS boot source.",
```
skatsaounis, line 17:
```suggestion
		Description: "Provides a resource to manage the MAAS boot source.",
```

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,126 @@
...
+func resourceBootSourceCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
+
+	bootsourceParams := entity.BootSourceParams{
+		KeyringData:     d.Get("keyring_data").(string),
+		KeyringFilename: d.Get("keyring_filename").(string),
+		URL:             d.Get("url").(string),
+	}
+
+	bootsource, err := client.BootSources.Create(&bootsourceParams)
```
skatsaounis, line 58: Based on the assumption that there i only one boot source, instead of creating a new one here, we have to fetch the pre-existing one and update its fields based on what the user has provided to Terraform

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,126 @@
...
 
 	"github.com/canonical/gomaasclient/client"
 	"github.com/canonical/gomaasclient/entity"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+func resourceMAASBootSource() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceMaasBootSourceRead,
```
skatsaounis, line 16: You are missing the part where you declare the CRU functions. Please add it

examples/resources/maas_boot_source/import.sh
```diff
@@ -0,0 +1,2 @@
+# Devices can be imported using their ID, e.g:
```
skatsaounis, line 1: Import is not valid for this resource, since Terraform assumes that there is only one boot source. Please remove it

examples/resources/maas_boot_source/resource.tf
```diff
@@ -0,0 +1,3 @@
+resource "maas_boot_source" "test_boot_source" {
+  description = "Test description"
```
skatsaounis, line 2: This one should explain how to set the URL pointing to candidate images

skatsaounis (changes requested):
I added a first set of suggestions. Please take a look and let's re-review the PR when you are ready. Nice effort so far :slightly_smiling_face:

maas/data_source_maas_boot_source.go
```diff
@@ -12,7 +12,7 @@ import (
 func dataSourceMaasBootSource() *schema.Resource {
 	return &schema.Resource{
 		ReadContext: dataSourceMaasBootSourceRead,
-		Description: "Provides a resource to fetch the MAAS boot source.",
+		Description: "Provides a resource to manage the MAAS boot source.",
```
skatsaounis, line 15: this one needs to be reverted. The data source is what Terraform is giving us to fetch resources. The initial message was correct.

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,146 @@
...
+
+	bootsource, err := getBootSource(client)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(fmt.Sprintf("%v", bootsource.ID))
+
+	// We create by changing the url to the chosen value
+	bootsourceParams := entity.BootSourceParams{
+		URL: d.Get("url").(string),
```
skatsaounis, line 64: what about the keyring fields? we should try to set them if any of them is specified by the user

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,146 @@
...
+
+func resourceBootSourceRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
+
+	bootsource, err := getBootSource(client)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(fmt.Sprintf("%v", bootsource.ID))
```
skatsaounis, line 83: This line is not needed. The Id is already set during the Create function's execution

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,146 @@
...
+	bootsourceParams := entity.BootSourceParams{
+		KeyringData:     d.Get("keyring_data").(string),
+		KeyringFilename: d.Get("keyring_filename").(string),
+		URL:             d.Get("url").(string),
+	}
+
+	if _, err := client.BootSource.Update(bootsource.ID, &bootsourceParams); err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(fmt.Sprintf("%v", bootsource.ID))
```
skatsaounis, line 117: This line is not needed. The Id is already set during the Create function's execution

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,146 @@
...
+func resourceBootSourceDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
+
+	bootsource, err := getBootSource(client)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(fmt.Sprintf("%v", bootsource.ID))
+
+	// We create by changing the url to the chosen value
```
skatsaounis, line 131:
```suggestion
	// We delete by changing the url to the default value
```

skatsaounis: But this is not enough. We also have to set the keyring file to the default one and also remove the keyring data if any

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,146 @@
...
+	bootsourceParams := entity.BootSourceParams{
+		URL: "http://images.maas.io/ephemeral-v3/stable/",
```
skatsaounis, line 133: `http://images.maas.io/ephemeral-v3/stable/` should be a [constant](https://go.dev/tour/basics/15)

maas/data_source_maas_boot_source.go
skatsaounis: In addition, the docs should be re-generated

skatsaounis (changes requested):
Another set of suggestions.

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,150 @@
...
+}
+
+func resourceBootSourceUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
+
+	bootsource, err := getBootSource(client)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(fmt.Sprintf("%v", bootsource.ID))
```
skatsaounis, line 109: this should not happen inside the update

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,150 @@
...
 	"fmt"
 
 	"github.com/canonical/gomaasclient/client"
 	"github.com/canonical/gomaasclient/entity"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+const defaultURL = "http://images.maas.io/ephemeral-v3/stable/"
+const defaultKeyring = "/snap/maas/current/usr/share/keyrings/ubuntu-cloudimage-keyring.gpg"
```
skatsaounis, line 14: this is a snap specific const. I think we should have a deb sepcific const too. In addition, if MAAS cannot provide via an API call whether it is a snap or deb installation, we may have to introduce a provider configuration. By doing so, users can tell the provider about the installation type and provider can be aware and decide things like this based on the value. The default value could be snap and the option would definitely be optional. We also have to add extra documentation to explain to users what is the purpose of this new provider config.

maas/resource_maas_boot_source_test.go
```diff
@@ -0,0 +1,101 @@
...
+		*bootSource = *gotBootSource
+
+		return nil
+	}
+}
+
+func testAccMAASBootSource(url string) string {
+	return fmt.Sprintf(`
+resource "maas_boot_source" "test" {
+	url = "%s"
```
skatsaounis, line 69: you should also set a keyring file here. Or just use the default one. Otherwise, the tests will fail since when you first create, the provider state will get the keyring information from MAAS during read. So, after the appliance with absent keyring file name, the provider will think that user wants to unset the keyring file name. Which is not the case.

maas/resource_maas_boot_source_test.go
```diff
@@ -0,0 +1,101 @@
...
+			continue
+		}
+
+		id, err := strconv.Atoi(rs.Primary.ID)
+		if err != nil {
+			return err
+		}
+		response, err := conn.BootSource.Get(id)
+		if err == nil {
+			if response.URL != default_url {
```
skatsaounis, line 90: we have to also check that the keyring file is back to the default and keyring data is also back to default which is empty ""

skatsaounis (commented):
Another set of suggestions

maas/resource_maas_boot_source.go
```diff
@@ -1,12 +1,150 @@
...
+			"created": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The creation time of the boot source.",
+			},
+			"keyring_data": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "The data on the keyring for the boot source.",
+				ConflictsWith: []string{"keyring_filename"},
```
skatsaounis, line 34: It seems that we should use ExactlyOneOf instead of ConflictsWith. Based on MAAS codebase, at least one of them must be set: https://github.com/canonical/maas/blob/master/src/maasserver/models/bootsource.py#L33-L48

```suggestion
				ExactlyOneOf: []string{"keyring_filename"},
```

maas/provider.go
```diff
@@ -30,6 +32,12 @@ func Provider() *schema.Provider {
 				Default:     "2.0",
 				Description: "The MAAS API version (default 2.0)",
 			},
+			"install_method": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     os.Getenv("MAAS_INSTALL_METHOD"),
```
skatsaounis:
```suggestion
				DefaultFunc: schema.EnvDefaultFunc("MAAS_INSTALL_METHOD", "snap"),
```

skatsaounis (changes requested):
This is the last set of suggestions. Please take a look. In addition, let's try to update the PR description with a detailed list of changes. Finally, since the `installation_method` is something new, please consider updating the `templates/index.md.tmpl` with some wording about it. Then `make generate_docs` is needed since `docs/index.md` is populated from the template.

When everything is done, we are good to go :rocket:

skatsaounis (approved):
LGTM +1

PR #258, opened 2025-02-06
feat: add resource and data source for MAAS zones

peterctl:
Managing MAAS availability zones through Terraform is a missing feature that I needed for a project.

Acceptance test output for the zone resource and datasource can be seen here:
```bash
$ make testacc TESTARGS="-run Zone"
TF_ACC=1 go test $(go list ./... | grep -v 'vendor') -v -run Zone -timeout 120m -parallel=4
?       terraform-provider-maas [no test files]
?       terraform-provider-maas/maas/testutils  [no test files]
=== RUN   TestAccDataSourceMaasZone_basic
=== PAUSE TestAccDataSourceMaasZone_basic
=== RUN   TestAccResourceMaasZone_basic
=== PAUSE TestAccResourceMaasZone_basic
=== CONT  TestAccDataSourceMaasZone_basic
=== CONT  TestAccResourceMaasZone_basic
--- PASS: TestAccDataSourceMaasZone_basic (2.61s)
--- PASS: TestAccResourceMaasZone_basic (3.37s)
PASS
ok      terraform-provider-maas/maas    3.386s
```


maas/resource_maas_zone.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	return nil
+}
+
+func resourceZoneCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
+
+	params := getZoneParams(d)
+	zone, err := findZone(client, params.Name)
```
skatsaounis, line 77: Trying to find if zone already exists in MAAS here costs an extra API call to MAAS. It is accepted to blindly try create the zone as the intention is that Terraform is managing it. If it already exists, MAAS will throw an error and this will be propagated to gomaasclient, then from the client to terraform provider and finally being presented to user as part of diagnostic errors. Then the user can import the zone to Terraform and continue its management with Terraform.

Let's simplify the function. Resource pool is again a good example as the implementation is more or less identical.

maas/resource_maas_zone.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	if zone == nil {
+		zone, err = client.Zones.Create(params)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+	}
+	d.SetId(zone.Name)
```
skatsaounis, line 87: As said earlier, let's save the database ID as a Terraform resource ID

maas/resource_maas_zone.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return resourceZoneUpdate(ctx, d, meta)
```
skatsaounis, line 89: Since the resource is just created, there is no need to try to modify with the same values that were used for its creation

```suggestion
	return resourceZoneRead(ctx, d, meta)
```

maas/resource_maas_zone.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func resourceZoneUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
+
```
skatsaounis, line 94: This is needed since we want to save the database ID. Then `id` should be given to `Zone.Update` instead of `d.Id()`.

```suggestion
id, err := strconv.Atoi(d.Id())
	if err != nil {
		return diag.FromErr(err)
	}
```

maas/resource_maas_zone.go
```diff
@@ -0,0 +1,142 @@
...
+		return diag.FromErr(err)
+	}
+
+	return resourceZoneRead(ctx, d, meta)
+}
+
+func resourceZoneDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
+
+	if err := client.Zone.Delete(d.Id()); err != nil {
```
skatsaounis, line 106: This can be simplified to

```suggestion
	return client.Zone.Delete(id))
```

given that id is calculated like the previous suggestion

maas/resource_maas_zone_test.go
```diff
@@ -0,0 +1,118 @@
...
+	description := "Test description"
+
+	checks := []resource.TestCheckFunc{
+		testAccMaasZoneCheckExists("maas_zone.test", &zone),
+		resource.TestCheckResourceAttr("maas_zone.test", "name", name),
+		resource.TestCheckResourceAttr("maas_zone.test", "description", description),
+	}
+
+	resource.ParallelTest(t, resource.TestCase{
+		Providers:    testutils.TestAccProviders,
```
skatsaounis, line 30: A precheck is missing here:

```go
PreCheck:     func() { testutils.PreCheck(t, nil) },
```

This function will check that `MAAS_API_URL` and `MAAS_API_KEY` are set before running the acceptance test.

maas/resource_maas_zone_test.go
```diff
@@ -0,0 +1,118 @@
...
+		CheckDestroy: testAccCheckMaasZoneDestroy,
+		ErrorCheck:   func(err error) error { return err },
+		Steps: []resource.TestStep{
+			{
+				Config: testAccMaasZone(name, description),
+				Check:  resource.ComposeTestCheckFunc(checks...),
+			},
+			// Test import using name
```
skatsaounis, line 38: This is also missing:

```go
                        // Test import using ID
			{
				ResourceName:      "maas_zone.test",
				ImportState:       true,
				ImportStateVerify: true,
			},
```

maas/resource_maas_zone_test.go
```diff
@@ -0,0 +1,118 @@
...
+		if !ok {
+			return fmt.Errorf("resource not found: %s\n %#v", rn, s.RootModule().Resources)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("resource id not set")
+		}
+
+		conn := testutils.TestAccProvider.Meta().(*client.Client)
+		gotZone, err := conn.Zone.Get(rs.Primary.ID)
```
skatsaounis, line 69: This is also needed if we are to save the database ID as the Terraform resource ID

```go
                id, err := strconv.Atoi(rs.Primary.ID)
		if err != nil {
			return err
		}
```

maas/resource_maas_zone_test.go
```diff
@@ -0,0 +1,118 @@
...
+	// is destroyed
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "maas_zone" {
+			continue
+		}
+
+		// Retrieve our maas_zone by referencing it's state ID for API lookup
+		response, err := conn.Zone.Get(rs.Primary.ID)
+		if err == nil {
+			if response != nil && response.Name == rs.Primary.ID {
```
skatsaounis, line 103:
```suggestion
			if response != nil && response.ID == rs.Primary.ID {
```

skatsaounis (changes requested):
Hi @peterctl 

Awesome work. Thank you for your first (I am looking forward to more) contribution to the provider. I tried to be exhaustive with comments to help you tackle the issues faster. Apart from 1-2 things that we can debate, rest of the suggested changes are needed.

Please take a look and let me know to re-review.

PS. Please include examples for data-source, resource and import and run `make generate_docs` to produce Markdown for the new resources. When you are done, please include the files to your PR too. Naming of directories and files is important for the tool to pick up the examples.

Reference:
- https://github.com/canonical/terraform-provider-maas/blob/master/examples/data-sources/maas_resource_pool/data-source.tf
- https://github.com/canonical/terraform-provider-maas/blob/master/examples/resources/maas_resource_pool/resource.tf
- https://github.com/canonical/terraform-provider-maas/blob/master/examples/resources/maas_resource_pool/import.sh

:beers:

peterctl:
Hi @skatsaounis,

Thank you for the thorough review! It's my first time working directly with Terraform providers so the feedback is very useful to me. I have applied most of the requested changes already, but I have an issue with setting the database ID as the Terraform ID instead of the name. The reason for this is that the gomaasclient will only accept names as arguments to `Get`, `Update` and `Destroy`. If I try using the numeric database ID, I get 404 errors. I wrote this quick snippet to show you what I get.

```go
// client_tests.go
package main

import (
        "fmt"
        "os"
        "terraform-provider-maas/maas"
)

func playground() error {
        config := maas.Config{
                APIURL: os.Getenv("MAAS_API_URL"),
                APIKey: os.Getenv("MAAS_API_KEY"),
                ApiVersion: "2.0",
        }

        client, err := config.Client()
        if err != nil {
                return err
        }

        zoneByName, errByName := client.Zone.Get("az1")
        zoneByID, errByID := client.Zone.Get("3")
        fmt.Printf("zoneByName: %+v\n", zoneByName)
        fmt.Printf("errByName: %+v\n", errByName)
        fmt.Printf("zoneByID: %+v\n", zoneByID)
        fmt.Printf("errByID: %+v\n", errByID)

        return nil
}

func main() {
        err := playground()
        fmt.Println(err)
}
```
```bash
$ go run client_playground.go
zoneByName: &{Name:az1 Description: ResourceURI:/MAAS/api/2.0/zones/az1/ ID:3}
errByName: <nil>
zoneByID: &{Name: Description: ResourceURI: ID:0}
errByID: ServerError: 404 Not Found (No Zone matches the given query.)
```

For clarity, the ID and name do match as can be seen in `zoneByName`.

If we want to use the database ID as the Terraform ID, I think we would need to rely on `findZone` to get the zone entity, then using `zone.Name` as the argument for the client, but this solution would add MAAS API calls any time we need to find the zone name from the database ID. For example:

```go
params := getZoneParams(d)
zone, err := getZone(d.Id())
if err != nil {
        return diag.FromErr(err)
}

zone, err = client.Zone.Update(zone.Name, params)
if err != nil {
        return diag.FromErr(err)
}
```

Please advise on how to proceed with name vs ID. Is there a better approach?

skatsaounis:
Hi @peterctl ,

The fact that MAAS API v2 is using the zone name instead of the database ID is problematic by itself. Since we cannot change the contract, it is what it is. With the v3 MAAS API we aim to solve this issue among others, but it is not there yet. As you spotted, the findZone is what you need here. We may add an extra API call, but it is a cheap one since zones response is small and we should expect something in the magnitude of tens or hundreds. What is the benefit though? Imagine that someone outside terraform, updated the zone name but there are already existing machines assigned to that zone. What you would expect from Terraform is to identify that change and try to update the same zone, instead of creating a new record. One of the weird behaviors of the API v2 is that you can use `/zones/zone_name/update?name=new_zone_name`, but again it is what it is.

Please proceed with the database ID as the Terraform resource ID.

In addition, we have to add a field of machines, like we do with tags so that we can assign the zone to machines with Terraform. I forgot to tell you in the first review. An alternative would be a new resource `maas_machine_zone_assignment` or `maas_zone_machine_assignment` or another alternative. I am undecided about which one is the best approach. But we can start with the machines field like tags to introduce a similar mechanism.

peterctl:
Hi @skatsaounis ,

I submitted the updated code using the Zone ID instead of the Name as the Terraform ID.

Regarding the `machines` field and associated options, I see a difference between Zones and Tags that affects this: Tag assignments are managed via the tag, while Zone assignments are managed via the machine. In the CLI (which is what I'm most familiar with) you set a tag via `maas user tag update-nodes $tag add=$nodes_to_add` and read tagged nodes via `maas user tag nodes $tag`. However, `maas user zone ...` does not have this functionality, but rather zones are assigned via `maas user machine update $machine zone=$zone`. This is currently reflected on both the `maas_machine` and `maas_vm_host_machine` resources, so there is already a method to set machine zones using the Terraform provider. Please let me know what you think.

skatsaounis (changes requested):
@peterctl thank you for implementing the changes. I think it is ok for now to keep the zone assignment as it is already implemented in the provider. I have added the proper label to run the system tests against your PR. If we get the green light I am more than happy to merge it :slightly_smiling_face: 

I have a last request from you. Please include examples on how to use the resource and datasource and how a zone can be imported. Then re-generate the docs with `make generate_docs`. Then commit the files to the PR.

You can draw inspiration from e.g., resource pools:

- https://github.com/canonical/terraform-provider-maas/tree/master/examples/resources/maas_resource_pool
- https://github.com/canonical/terraform-provider-maas/tree/master/examples/data-sources/maas_resource_pool

To let the files being included to the markdown, please pay attention to directory and file names.

peterctl:
Hey @skatsaounis ,

I included the examples into the docs and pushed them to the PR. Please let me know what you think :smile:

examples/data-sources/maas_zone/data-source.tf
```diff
@@ -0,0 +1,8 @@
+data "maas_zone" "test_zone" {
```
skatsaounis, line 1:
```suggestion
resource "maas_zone" "test_zone" {
```

skatsaounis (changes requested):
We are (almost) good to go. Please fix the example pointed by the suggestion and regenerate the docs afterwards. Then, we are good to go :rocket:

peterctl:
@skatsaounis example fixed :+1:

skatsaounis:
@peterctl you forgot to regenerate the docs, and the CI is complaining. Please do so that we can merge it :)

peterctl:
Sorry, I accepted the comment's suggestion and forgot about regenerating the docs. I just pushed the updated commit and rebased the branch on top of the current master.

maas/data_source_maas_zone.go
```diff
@@ -0,0 +1,49 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The zone's name.",
+			},
+		},
+	}
+}
+
+func dataSourceZoneRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
```
SK1Y101, line 33: We recently swapped this out with a clientConfig object, you'll need to swap this out with:
`client := meta.(*ClientConfig).Client`

maas/resource_maas_zone.go
```diff
@@ -0,0 +1,142 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The name of the new zone.",
+			},
+		},
+	}
+}
+
+func resourceZoneRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
```
SK1Y101, line 48: `client := meta.(*ClientConfig).Client`

maas/resource_maas_zone.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if err := setTerraformState(d, tfstate); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func resourceZoneCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
```
SK1Y101, line 70: `client := meta.(*ClientConfig).Client`

maas/resource_maas_zone.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(fmt.Sprintf("%v", zone.ID))
+
+	return resourceZoneRead(ctx, d, meta)
+}
+
+func resourceZoneUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
```
SK1Y101, line 83: `client := meta.(*ClientConfig).Client`

maas/resource_maas_zone.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(fmt.Sprintf("%v", zone.ID))
+
+	return resourceZoneRead(ctx, d, meta)
+}
+
+func resourceZoneDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client.Client)
```
SK1Y101, line 100: `client := meta.(*ClientConfig).Client`

maas/resource_maas_zone_test.go
```diff
@@ -0,0 +1,137 @@
...
+		rs, ok := s.RootModule().Resources[rn]
+		if !ok {
+			return fmt.Errorf("resource not found: %s\n %#v", rn, s.RootModule().Resources)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("resource id not set")
+		}
+
+		conn := testutils.TestAccProvider.Meta().(*client.Client)
```
SK1Y101, line 74: `conn := testutils.TestAccProvider.Meta().(*maas.ClientConfig).Client`

maas/resource_maas_zone_test.go
```diff
@@ -0,0 +1,137 @@
...
+resource "maas_zone" "test" {
+	name        = "%s"
+	description = "%s"
+}
+`, name, description)
+}
+
+func testAccCheckMaasZoneDestroy(s *terraform.State) error {
+	// retrieve the connection established in Provider configuration
+	conn := testutils.TestAccProvider.Meta().(*client.Client)
```
SK1Y101, line 97: `conn := testutils.TestAccProvider.Meta().(*maas.ClientConfig).Client`

maas/resource_maas_zone.go
```diff
@@ -0,0 +1,142 @@
...
+func resourceMaasZone() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Provides a resource to manage MAAS zones.",
+		CreateContext: resourceZoneCreate,
+		ReadContext:   resourceZoneRead,
+		UpdateContext: resourceZoneUpdate,
+		DeleteContext: resourceZoneDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: func(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+				client := meta.(*client.Client)
```
SK1Y101, line 22: `client := meta.(ClientConfig).Client`

SK1Y101 (changes requested):
Looks good, but does need a small syntax change to match the master branch

peterctl:
@SK1Y101 thanks for the pointers :smiley: I updated the PR to reflect your requested changes.

SK1Y101 (approved):

SK1Y101:
@peterctl Awesome, thanks for the changes. I've sent it back off for testing, that'll be completed somewhen soon

skatsaounis (approved):
LGTM +1 :rocket:

Issue #261, opened 2025-02-11
ServerError: 400 `name` error when creating interface


SK1Y101:
Running the acceptance tests on a MAAS 3.6 instance yields the following error message:
```bash
Error: ServerError: 400 Bad Request ({"name": ["The interface name must be 1 to 15 characters long and can only contain alphanumeric characters, hyphens ('-'), underscores ('_'), periods ('.'), and colons (':')."]})
```
For the tests `resource_maas_network_interface_bond_test`, `resource_maas_network_interface_physical_test`, and `resource_maas_network_interface_bridge_test`.

It seems MAAS 3.6 introduced a new restriction on interface names, which further restricts the allowed interface names.

We assign interface names from a prefex, ie: 
```go
name := acctest.RandomWithPrefix("tf-network-interface-bond")
```
Which is already too long. We should refactor this to reduce length

PR #262, opened 2025-02-11
refactor: Reduce testing interface name length

SK1Y101:
Addresses #261 
- Replaces `RandomWithPrefix` with a formatted string whose length is below the 15 character limit, using single digit `RandIntRange` as a suffix instead.


skatsaounis (approved):
LGTM +1

Issue #254, opened 2025-02-04
partitions in block devices are created in different order for each execution


pescobar:
Hi,

We are deploying multiple machines using the same terraform code and a `for_each` loop and we noticed that the resulting partitions layout is different between machines even if the terraform code used to generate the partition layout is the same. 

Consecutive executions of the same code which should return `No changes. Your infrastructure matches the configuration`  tries to update the partition layout for already deployed machines which triggers the error `Cannot update block device because the machine is not Ready.`

This is the terraform code we apply to all our nodes:

```
resource "maas_block_device" "sda" {
  for_each       = var.compute_nodes
  machine        = maas_machine.compute_nodes[each.key].id
  name           = "sda"
  is_boot_device = true
  id_path        = "/dev/sda"
  size_gigabytes = 959
  tags = [
    "ssd",
  ]

  partitions {
    size_gigabytes = 50
    fs_type        = "ext4"
    label          = "media"
    mount_point    = "/"
  }

  partitions {
    size_gigabytes = 50
    fs_type        = "ext4"
    mount_point    = "/var"
  }

  partitions {
    size_gigabytes = 50
    fs_type        = "ext4"
    mount_point    = "/var/log"
  }

  partitions {
    size_gigabytes = 50
    fs_type        = "ext4"
    mount_point    = "/var/adm"
  }

  partitions {
    size_gigabytes = 50
    fs_type        = "ext4"
    mount_point    = "/var/tmp"
  }

  partitions {
    size_gigabytes = 50
    fs_type        = "ext4"
    mount_point    = "/var/log/audit"
  }

  partitions {
    size_gigabytes = 50
    fs_type        = "ext4"
    mount_point    = "/tmp"
  }

}
```

And after applying this same exact code to two identical machines we end up with this partition layout in each of the machines. Notice that the file systems order in the first column differs between the two nodes.

![Image](https://github.com/user-attachments/assets/31893bf6-b043-431d-b8b7-636d83532b9a)

![Image](https://github.com/user-attachments/assets/97d764ab-75bd-4e05-a7c1-8bae3909303e)

 This order also changes in consecutive executions for already deployed machines and opentofu tries to update deployed machines e.g. see the plan output

```
  # module.compute_gigabyte_H262_Z63_00_128c_1tb.maas_block_device.sda["compute06"] will be updated in-place
  ~ resource "maas_block_device" "sda" {
        id             = "113"
        name           = "sda"
        tags           = [
            "ssd",
        ]
[...]

PR #263, opened 2025-02-12
fix: determine the correct partition order

skatsaounis:
Sort the block device partitions based on their database ID since MAAS API v2 does not provide any guarantee on the order they are returned.

Resolves: #254


cpg1111 (approved):
LGTM

PR #264, opened 2025-02-13
feat: Add boot source selection

SK1Y101:
Add terraform data sources for boot-source-selection.
Also adds CRUD operations and import operation for boot-source-selection.

Note: Behaviour for the default commissioning release is different to others. On a delete, MAAS will throw an error that will be reported, as you cannot delete the default commissioning release.


maas/data_source_maas_boot_source_selection.go
```diff
@@ -0,0 +1,118 @@
...
+		"subarches":      bootsourceselection.Subarches,
+	}
+	if err := setTerraformState(d, tfState); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func getBootSourceSelection(client *client.Client, boot_source_id int, id int) (*entity.BootSourceSelection, error) {
```
skatsaounis, line 85: please move to resource_maas_boot_source_selection.go as it is used there only.

maas/data_source_maas_boot_source_selection.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if bootsourceselection == nil {
+		return nil, fmt.Errorf("boot source selection (%s %s) was not found", os, release)
+	}
+	return bootsourceselection, nil
+}
+
+func findBootSourceSelection(client *client.Client, boot_source_id int, os string, release string) (*entity.BootSourceSelection, error) {
```
skatsaounis, line 107: please move to resource_maas_boot_source_selection.go as it is used there only.

maas/data_source_maas_boot_source_selection.go
```diff
@@ -0,0 +1,118 @@
...
+				Description: "The list of subarches for this resource",
+			},
+		},
+	}
+}
+
+func dataSourceMaasBootSourceSelectionRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*ClientConfig).Client
+
+	bootsource, err := getBootSource(client)
```
skatsaounis, line 59: please use the ID provided by the user at the `boot_source` field given that it is renamed

maas/data_source_maas_boot_source_selection.go
```diff
@@ -0,0 +1,118 @@
...
+		Description: "Provides a resource to fetch a MAAS boot source selection.",
+
+		Schema: map[string]*schema.Schema{
+			"arches": {
+				Type:        schema.TypeSet,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Description: "The architecture list for this resource",
+			},
+			"boot_source_id": {
```
skatsaounis, line 25: Please rename to `boot_source` and in the description please use  `the boot source database ID` or something similar to inform the practitioner about the field

maas/data_source_maas_boot_source_selection.go
```diff
@@ -0,0 +1,118 @@
...
+				Description: "The label lists for this resource",
+			},
+			"os": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The Operating system for this resource",
+			},
+			"release": {
+				Type:        schema.TypeString,
+				Optional:    true,
```
skatsaounis, line 43:
```suggestion
				Required:    true,
```

maas/data_source_maas_boot_source_selection.go
```diff
@@ -0,0 +1,118 @@
...
+		Schema: map[string]*schema.Schema{
+			"arches": {
+				Type:        schema.TypeSet,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Description: "The architecture list for this resource",
+			},
+			"boot_source_id": {
+				Type:        schema.TypeInt,
+				Computed:    true,
```
skatsaounis, line 27:
```suggestion
				Required:    true,
```

maas/resource_maas_boot_source_selection.go
```diff
@@ -0,0 +1,227 @@
...
+		Schema: map[string]*schema.Schema{
+			"arches": {
+				Type:        schema.TypeSet,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Description: "The architecture list for this resource",
+			},
+			"boot_source_id": {
+				Type:        schema.TypeInt,
+				Computed:    true,
```
skatsaounis, line 32:
```suggestion
				Required:    true,
```

maas/resource_maas_boot_source_selection.go
```diff
@@ -0,0 +1,227 @@
...
+				Description: "The list of subarches for this resource",
+			},
+		},
+	}
+}
+
+func resourceBootSourceSelectionCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*ClientConfig).Client
+
+	bootsource, err := getBootSource(client)
```
skatsaounis, line 64: This is not needed. We will get the boot_source ID from the user.

maas/resource_maas_boot_source_selection.go
skatsaounis: And also rename to `boot_source`

maas/resource_maas_boot_source_selection.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	d.SetId((fmt.Sprintf("%v", bootsourceselection.ID)))
+
+	return resourceBootSourceSelectionRead(ctx, d, meta)
+}
+
+func resourceBootSourceSelectionRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*ClientConfig).Client
+
+	bootsource, err := getBootSource(client)
```
skatsaounis, line 89: this is provided by the user. It is not needed

maas/resource_maas_boot_source_selection.go
```diff
@@ -0,0 +1,227 @@
...
+	id, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	bootsourceselection, err := getBootSourceSelection(client, bootsource.ID, id)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
```
skatsaounis, line 103: Maybe we need to include this `d.SetId((fmt.Sprintf("%v", bootsourceselection.ID)))`

maas/resource_maas_boot_source_selection.go
```diff
@@ -0,0 +1,227 @@
...
+	if err := setTerraformState(d, tfState); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+func resourceBootSourceSelectionUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*ClientConfig).Client
+
+	bootsource, err := getBootSource(client)
```
skatsaounis, line 121: this is not needed. It is provided by the user

maas/resource_maas_boot_source_selection.go
```diff
@@ -0,0 +1,227 @@
...
+		return diag.FromErr(err)
+	}
+
+	return resourceBootSourceSelectionRead(ctx, d, meta)
+}
+
+func resourceBootSourceSelectionDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*ClientConfig).Client
+
+	bootsource, err := getBootSource(client)
```
skatsaounis, line 154: This is not needed. It is provided by the user

maas/resource_maas_boot_source_selection.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	err = json.Unmarshal(default_release_bytes, &default_release)
+	if err != nil {
+		return nil, err
+	}
+
+	// Only then fetch the default bootsource that refers to them
+
+	bootsource, err := getBootSource(client)
```
skatsaounis, line 221: This is not needed. Actually, the check should be on the `d.os == "ubuntu" && d.release == commissioning_distro_series`. There is no need to fetch the default selection from MAAS.

maas/resource_maas_boot_source_selection.go
```diff
@@ -0,0 +1,227 @@
...
+	id, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	bootsourceselection, err := getBootSourceSelection(client, bootsource.ID, id)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
```
skatsaounis, line 168: Here if we are not the default OS, we simply delete. If we are the default OS we are updating.

maas/resource_maas_boot_source_selection.go
```diff
@@ -0,0 +1,227 @@
...
+)
+
+func resourceMAASBootSourceSelection() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Provides a resource to manage a MAAS boot source selection.",
+		CreateContext: resourceBootSourceSelectionCreate,
+		ReadContext:   resourceBootSourceSelectionRead,
+		UpdateContext: resourceBootSourceSelectionUpdate,
+		DeleteContext: resourceBootSourceSelectionDelete,
+
```
skatsaounis, line 22: We are missing an import function. Content: https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/import

skatsaounis (changes requested):
The PR looks good so far. I tried to capture the big things that stand out in a first review round. Please take a look, and as always, after all changes, let's regenerate the docs. For the docs part, we will also need an import.sh file

skatsaounis (changes requested):
We are good to merge. What is missing though is an example of the data source. Please add it and regenerate the docs.

examples/data-sources/maas_boot_source_selection/data-source.tf
```diff
@@ -0,0 +1,8 @@
+data "maas_boot_source" "default" {}
+
+data "maas_boot_source_selection" "default" {
+  boot_source = maas_boot_source.default.id
```
skatsaounis, line 4: This one should be fixed, and the docs should be regenerated.

```suggestion
  boot_source = data.maas_boot_source.default.id
```

examples/resources/maas_boot_source_selection/import.sh
```diff
@@ -0,0 +1,2 @@
+# A bond network interface can be imported using the boot source ID and ID. e.g.
```
skatsaounis, line 1: Including doc regeneration.

```suggestion
# A boot resource selection can be imported using the boot source ID and the boot source selection ID. e.g.
```

skatsaounis (changes requested):
Please fix the examples and the documentation. In addition, update the PR description to describe accurately the changes of the PR. This time, let's first reach Approved state and run the system tests afterward.

skatsaounis (approved):
LGTM +1

skatsaounis:
relates to: #128

Issue #259, opened 2025-02-10
ServerError: 400 Bad Request when creating link resource for bridge resource with VLAN as parent


kolovo:
**Description:**

I am encountering a `400 Bad Request` error when trying to create a link resource for a bridge resource that has a VLAN as its parent in the MAAS Terraform provider. The error message states:

```
ServerError: 400 Bad Request ({"vlan": ["VLAN interface requires connection to a VLAN."]})
```

Based on the documentation, the following Terraform configuration should work for attaching a VLAN to a physical interface (`eth0`) and then attaching the VLAN to a bridge:

```hcl
resource "maas_network_interface_vlan" "example" {
  machine = "test-vm1"
  parent  = "eth0"
  vlan    = 30
  fabric  = "fabric-0"
}

resource "maas_network_interface_bridge" "example" {
  machine = "test-vm1"
  name    = "br30"
  parent  = maas_network_interface_vlan.example.id
  vlan    = 30 (same result even when not setting this)
}

resource "maas_network_interface_link" "virsh_vm1_nic1" {
  machine           = "test-vm1"
  network_interface = maas_network_interface_bridge.example.id
  subnet            = "10.0.30.0/24"
  mode              = "STATIC"
  ip_address        = "10.0.30.155"
  default_gateway   = true
}
```

However, the error suggests that the VLAN interface is not correctly associated with a VLAN.

### Steps to Reproduce:
1. Apply the above Terraform configuration.
2. Observe the error message returned by MAAS.

### Expected Behavior:
The Terraform configuration should successfully create:
- A VLAN interface on `eth0`.
- A bridge interface on top of the VLAN.
- A link resource that assigns an IP to the bridge.

### Actual Behavior:
Terraform fails with the following error:

```
ServerError: 400 Bad Request ({"vlan": ["VLAN interface requires connection to a VLAN."]})
```

### Additional Information:
- Terraform Provider version: 2.4.0
- MAAS Version:3.5.3

Am I missing something in the configuration, or is this a bug in the provider?

---
**Request:** Please confirm if this is an expected behavior or a bug. If it is an expected behavior, it would be helpful to update the documentation with proper guidance on setting up VLANs with bridges correctly.

skatsaounis:
Hi @kolovo 

[...]

PR #268, opened 2025-02-18
fix: properly unlink existing links before creating a link

skatsaounis:
- Unlink all existing links of an interface before create a new link, instead of disconnect
- Add a new attribute to vlan interface to set a name or calculate it from MAAS
- Add acceptance tests for `maas_interface_link`
- Make `vlan` field of a vlan interface a mandatory field with force re-creation, since it was accidentally marked as optional

Resolves: #259


SK1Y101 (approved):
Nice!

PR #272, opened 2025-02-21
fix: IP ranges cannot detect changes

tobiasdemendonca:
## Description of changes

Adds IP range fields to IP range read function. This fixes https://github.com/canonical/terraform-provider-maas/issues/243, where Terraform wasn't keeping track of changes to the start or end IP addresses of existing IP ranges.

Issue: https://github.com/canonical/terraform-provider-maas/issues/243

## Checklist

### DIFF
--- a/maas/resource_maas_subnet_ip_range_test.go
+++ b/maas/resource_maas_subnet_ip_range_test.go
@@ -87,6 +87,13 @@ func TestAccResourceMaasSubnetIPRange_basic(t *testing.T) {
 				ImportState:       true,
 				ImportStateVerify: true,
 			},
+			// Test import using start_ip:end_ip format
+			{
+				ResourceName:      ipRangeAttrName,
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateId:     fmt.Sprintf("%s:%s", ipStartMod, ipEndMod),
+			},
 		},
 	})
 }


tobiasdemendonca:
Have put this in draft form because I need to add some steps that test the change behaviour

maas/resource_maas_subnet_ip_range_test.go
```diff
@@ -0,0 +1,186 @@
...
+					subnet_name,
+					range_type,
+					comment,
+					ipStart,
+					ipEnd,
+				),
+				Check: resource.ComposeTestCheckFunc(checks...),
+			},
+			// Test if resource drift is detected by modifying the IP range using the
+			// GO MAAS client, and then checking if the state is updated correctly on
```
skatsaounis, line 57: nit:

```suggestion
			// Go MAAS client, and then checking if the state is updated correctly on
```

maas/resource_maas_subnet_ip_range_test.go
```diff
@@ -0,0 +1,186 @@
...
+					// Update the IP range to changed values that should be read into state
+					_, err := client.IPRange.Update(ipRange.ID, &params)
+					if err != nil {
+						t.Fatalf("Failed to update IP range: %s", err)
+					}
+				},
+				RefreshState:       true,
+				ExpectNonEmptyPlan: true,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(ipRangeAttrName, "start_ip", ipStartMod),
```
skatsaounis, line 78: since the test is modifying everything, let's also test all the other fields to confirm that they are also modified. Checking that subnet is the same is optional so feel free to check it or not.

maas/resource_maas_subnet_ip_range_test.go
```diff
@@ -0,0 +1,186 @@
...
+				),
+			},
+			// Test import
+			{
+				ResourceName:      ipRangeAttrName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
```
skatsaounis, line 87: Since you have added a test for the import, please also include one that is testing the import with the `start_ip:end_ip` e.g., `10.77.77.2:10.77.77.60`

skatsaounis: This test is modifying the resource first, so I would expect that the import should suceed with the modified start and end IPs.

maas/resource_maas_subnet_ip_range_test.go
```diff
@@ -0,0 +1,186 @@
...
+func testAccSubnetIPRangeExampleResource(
+	name_subnet string,
+	range_type string,
+	comment string,
+	start_ip string,
+	end_ip string,
+) string {
+	// A subnet is required to create an IP range
+	return fmt.Sprintf(`
+		resource "maas_subnet" "test_subnet" {
```
skatsaounis, line 130: This is still not properly formatted Terraform code. For example, gateway_ip and dns_servers `=` are not aligned with `=` symbols of the other fields. In addition, the valid Terraform indentation is 2 spaces. Finally, the string should start at the beginning of the line and can disrespect the Go indentation due to the usage of `fmt.Sprintf`. My advice for your future tests is to first write your Terraform in a .tf file, then ask your Terraform binary or IDE to format with `terraform fmt`, then paste inside the acceptance test function.

skatsaounis (changes requested):
This is a very well written PR and a great attempt. I tried to thoroughly review, so I have also included some nits. Apart from that, please take a look at the acceptance test comments. I am more than happy to approve when you take care of them.

NOTE: There is nothing stopping us to end up with different acceptance tests vs one basic. In fact, this is the proper way in order to test the resource use cases in isolation: i) a basic CRUD test, ii) a refresh state test (this is to confirm your fix is valid) iii) importing with all possible ways. However, since the acceptance testing in the provider is not so mature, I am fine to keep a single test but one that is also testing the missing import use case. :slightly_smiling_face:

tobiasdemendonca:
@skatsaounis thanks very much for your thorough review! I've made all changes youve suggested.

skatsaounis (approved):
LGTM +1

PR #274, opened 2025-02-25
feat: Add boot resource operations

SK1Y101:
## Description of changes

Extends the `maas_boot_source_selection` resource so that after the selection is applied, Terraform is requesting an import and is expecting it to finish. By doing so, the existence of `maas_boot_source_selection` guarantees the availability of the images related to the selection

Resolves: #128 

## Checklist


maas/resource_maas_boot_resource.go
```diff
@@ -0,0 +1,174 @@
...
+			},
+		},
+	}
+}
+
+func resourceBootResourcesCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*ClientConfig).Client
+
+	for isImporting(client) {
+		time.Sleep(5 * time.Second)
```
skatsaounis, line 44: let's convert this and any other sleep to `retry.StateChangeConf` and `stateConf.WaitForStateContext`. We already have an example at the `func waitForMachineStatus`

maas/provider.go
```diff
@@ -52,6 +52,7 @@ func Provider() *schema.Provider {
 			},
 		},
 		ResourcesMap: map[string]*schema.Resource{
+			"maas_boot_resource":              resourceMAASBootResources(),
```
skatsaounis, line 55:
```suggestion
			"maas_boot_resources":              resourceMAASBootResources(),
```

maas/resource_maas_boot_resource.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	for _, bootselection := range bootselections {
+		bootselection, err := getBootSourceSelection(client, d.Get("boot_source").(int), bootselection.(int))
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		// check the selection has it's resource created
+		if _, exists := resourceMap[fmt.Sprintf("%s/%s", bootselection.OS, bootselection.Release)]; !exists {
+			return diag.Errorf("Boot Resource missing for %s/%s", bootselection.OS, bootselection.Release)
```
skatsaounis, line 68: Should we re-trigger import and wait? Maybe not.

maas/resource_maas_boot_resource.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+
+	return nil
```
skatsaounis, line 72: what about running read here?

skatsaounis: Or what about using the same function for both read and update? (that is calling read in the end)

maas/resource_maas_boot_resource.go
SK1Y101: I think it's enough to tell the user there was an issue, and they can modify the plan and trigger an import

maas/resource_maas_boot_resource.go
```diff
@@ -0,0 +1,174 @@
...
+	client := meta.(*ClientConfig).Client
+
+	for isImporting(client) {
+		time.Sleep(5 * time.Second)
+	}
+
+	return resourceBootResourcesRead(ctx, d, meta)
+}
+
+func resourceBootResourcesDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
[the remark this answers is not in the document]
SK1Y101: sorry, out of the loop, wdym?

maas/resource_maas_boot_source_selection.go
```diff
@@ -143,12 +171,21 @@ func resourceBootSourceSelectionUpdate(ctx context.Context, d *schema.ResourceDa
...
+		return diag.FromErr(err)
+	}
+
 	return resourceBootSourceSelectionRead(ctx, d, meta)
 }
 
 func resourceBootSourceSelectionDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*ClientConfig).Client
 
+	err := awaitImportComplete(client)
```
skatsaounis, line 184: I think that this import is not needed here. Instead, we may have to run it after deleting the selection so that MAAS can clear the boot resources of this selection.

maas/data_source_maas_boot_resource_test.go
```diff
@@ -0,0 +1,173 @@
...
+	os      = maas_boot_source_selection.test.os
+	release = maas_boot_source_selection.test.release
+}
+`, testAccMAASBootSourceSelection(os, release, arches, subarches, labels))
+}
+
+func testAccCheckDataSourceMaasBootResourcesDestroy(s *terraform.State) error {
+	// retrieve the connection established in Provider configuration
+	conn := testutils.TestAccProvider.Meta().(*maas.ClientConfig).Client
+	if err := awaitImportComplete(conn); err != nil {
```
skatsaounis, line 72: If everything works as expected, when we enter this function, the import has finished. We should remove this awaitImportComplete to avoid hiding issues with the implementation.

maas/resource_maas_boot_resource.go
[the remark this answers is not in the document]
skatsaounis: never mind, I am not sure either about my initial idea after all the changes. I am resolving it for now.

examples/resources/maas_boot_resources/resource.tf
```diff
@@ -0,0 +1,26 @@
...
+}
+
+resource "maas_boot_resources" "test" {
+  boot_source = maas_boot_source.test_boot_source.id
+
+  boot_source_selections = [
+    maas_boot_source_selection.jammy_test.id,
+    maas_boot_source_selection.noble_test.id,
+  ]
+}
```
skatsaounis, line 26: nit: let's add a new line before the EOF.

maas/data_source_maas_boot_resource.go
```diff
@@ -0,0 +1,108 @@
...
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "The name of this resource.",
+						},
+						"subarches": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "The subarches for this resource.",
+						},
+						"type": {
```
skatsaounis, line 43: Type can be either synced or uploaded (custom). Since we only care for synced, what about not keeping track of it with Terraform at all? This one requires changes to couple of places (docs included) but it can simplify the resources. In case we want to also support custom, we need to keep it but update our logic. I favor the synced only option but it's up to you

maas/data_source_maas_boot_resource_test.go
```diff
@@ -0,0 +1,144 @@
...
+		},
+	})
+}
+
+func testAccDataSourceMaasBootResources(os string, release string, arches []string, subarches []string, labels []string) string {
+	return fmt.Sprintf(`
+%s
+
+data "maas_boot_resources" "test" {
+	os      = maas_boot_source_selection.test.os
```
skatsaounis, line 60: nit: what about using 2 spaces for tabs?

maas/data_source_maas_boot_resource_test.go
```diff
@@ -0,0 +1,144 @@
...
+				return fmt.Errorf("boot source selection (%s) was unexpectedly found on deleted resource", this_name)
+			}
+		}
+		return nil
+	}
+
+	return nil
+}
+
+func findBootSourceSelection(client *client.Client, boot_source int, os string, release string) (*entity.BootSourceSelection, error) {
```
skatsaounis, line 133: what about moving this to `resource_maas_boot_resource.go` like we do with others?

maas/data_source_maas_boot_source_selection.go
```diff
@@ -56,6 +56,11 @@ func dataSourceMaasBootSourceSelection() *schema.Resource {
 func dataSourceMaasBootSourceSelectionRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*ClientConfig).Client
 
+	err := awaitImportComplete(client)
```
skatsaounis, line 59: is there a reason to wait for the import to be completed? the selections are available instantly. In addition, if the Terraform practitioner wants to ensure that the boot resources are available, they have to use the dedicated data_source or resource for them.

maas/resource_maas_boot_resource.go
```diff
@@ -0,0 +1,330 @@
...
+func ResourceMaasBootResources() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Provides a resource to manage MAAS bootresources.",
+		CreateContext: resourceBootResourcesCreate,
+		ReadContext:   resourceBootResourcesRead,
+		UpdateContext: resourceBootResourcesUpdate,
+		DeleteContext: resourceBootResourcesDelete,
+
+		Schema: map[string]*schema.Schema{
+			"boot_source": {
```
skatsaounis, line 26: Question: Since we want to hide the fact that we can have many boot_sources in the database, what is the benefit for exposing it here? How can this be useful for the Terraform practitioner?

maas/resource_maas_boot_resource.go
```diff
@@ -0,0 +1,330 @@
...
+	d.SetId(fmt.Sprintf("%v", bootsource.ID))
+
+	var selectionSet []int
+	for _, bootselection := range bootselections {
+		bootselection, err := getBootSourceSelection(client, bootsource.ID, bootselection.(int))
+		if err != nil {
+			return diag.Errorf("error fetching boot selection %v: %s", bootselection, err)
+		}
+
+		// check the selection has it's resource created
```
skatsaounis, line 77: nit:
```suggestion
		// check the selection has its resource created
```

maas/resource_maas_boot_resource.go
```diff
@@ -0,0 +1,330 @@
...
+		return diag.Errorf("error fetching synced boot resources: %v", err)
+	}
+	resourceMap := make(map[string]struct{})
+	for _, res := range resources {
+		resourceMap[res.Name] = struct{}{}
+	}
+
+	bootselections := d.Get("boot_source_selections").(*schema.Set).List()
+
+	if len(bootselections) == 0 {
```
skatsaounis, line 60: let's also check for unique ids?

maas/resource_maas_boot_resource.go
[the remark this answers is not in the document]
skatsaounis: Ok I remember now. In the Terraform delete, we should do nothing on MAAS. Remember that we use the `maas_boot_resources` resource to make sure that the import is done or not. The deletion of boot source selections is not part of this. It can happen with `maas_boot_source_selection` if done by Terraform, or by the user outside of Terraform. So it has to be a noop here. `d.SetId("")` only or something similar.

maas/resource_maas_boot_resource.go
```diff
@@ -0,0 +1,330 @@
...
+			return diag.Errorf("error fetching boot selection %v: %s", bootselection, err)
+		}
+
+		// check the selection has it's resource created
+		if _, exists := resourceMap[fmt.Sprintf("%s/%s", bootselection.OS, bootselection.Release)]; !exists {
+			return diag.Errorf("Boot Resource missing for %s/%s", bootselection.OS, bootselection.Release)
+		}
+		selectionSet = append(selectionSet, bootselection.ID)
+	}
+	if err := d.Set("boot_source_selections", unique(selectionSet)); err != nil {
```
skatsaounis, line 83: if unique is ensured by validating user's input, then this unique converter is not needed.

maas/data_source_maas_boot_resource.go
```diff
@@ -0,0 +1,108 @@
+package maas
```
skatsaounis, line 1: name of the file should change to plural

maas/data_source_maas_boot_resource_test.go
```diff
@@ -0,0 +1,144 @@
+package maas_test
```
skatsaounis, line 1: name of the file should change to plural

maas/resource_maas_boot_resource.go
```diff
@@ -0,0 +1,330 @@
+package maas
```
skatsaounis, line 1: name of the file should change to plural

maas/resource_maas_boot_resource_test.go
```diff
@@ -0,0 +1,191 @@
+package maas_test
```
skatsaounis, line 1: name of the file should change to plural

maas/resource_maas_boot_resource_test.go
```diff
@@ -0,0 +1,191 @@
...
+		},
+	})
+}
+
+func testAccMaasBootResources() string {
+	return `
+data "maas_boot_source" "test" {}
+
+resource "maas_boot_source_selection" "mantic" {
+    boot_source = data.maas_boot_source.test.id
```
skatsaounis, line 51: nit: tab size should be 2 spaces

maas/resource_maas_boot_source_selection.go
```diff
@@ -77,6 +86,11 @@ func resourceBootSourceSelectionImport(ctx context.Context, d *schema.ResourceDa
 func resourceBootSourceSelectionCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*ClientConfig).Client
 
+	err := awaitImportComplete(client)
```
skatsaounis, line 89: Question: Please remind me why we need to trigger import inside the boot source selection Terraform CRUD operations? Since the only way to guarantee to the user that the resources are available, they have to use `maas_boot_resource`, then why we need to add it to selections too?

skatsaounis (changes requested):

maas/resource_maas_boot_resource.go
SK1Y101: We use it equally for the selections, or are you suggesting we remove that entirely and just fetch the boot source each time, rather than reading from the terraform?

maas/resource_maas_boot_source_selection.go
SK1Y101: Oh, it's theoretically possible to use terraform to handle boot images without also using boot resources, so i added import checks to selections, as it's impossible to use boot resources without selections.

maas/resource_maas_boot_source_selection.go
```diff
@@ -190,3 +235,77 @@ func getBootSourceSelection(client *client.Client, bootSource int, id int) (*ent
...
+			if resource.Name == fmt.Sprintf("%s/%s", os, release) {
+				archesFound.Insert(strings.Split(resource.Architecture, "/")[0])
+
+				resourceDetails, err := client.BootResource.Get(resource.ID)
+				if err != nil {
+					return retry.NonRetryableError(err)
+				}
+
+				for _, resourceSset := range resourceDetails.Sets {
+					if !resourceSset.Complete {
```
tobiasdemendonca, line 268: typo 
```suggestion
				for _, resourceSet := range resourceDetails.Sets {
					if !resourceSet.Complete {
```

maas/resource_maas_boot_source_selection.go
```diff
@@ -190,3 +235,77 @@ func getBootSourceSelection(client *client.Client, bootSource int, id int) (*ent
 
 	return bootSourceSelection, nil
 }
+
+func awaitImportComplete(client *client.Client, os string, release string, arches []string) error {
+	if err := client.BootResources.Import(); err != nil {
+		return err
+	}
+
+	timeout := 40 * time.Minute
```
tobiasdemendonca, line 244: Nit: Were we going to make this user setable, or did you decide against this? It may not be worth the effort, I just want to understand. The same applies in delete. 

If this isn't user setable, is this meant to be the same as the delete timeout? If so, do you think we need these to be constants to avoid magic numbers?

maas/resource_maas_boot_source_selection.go
```diff
@@ -190,3 +235,77 @@ func getBootSourceSelection(client *client.Client, bootSource int, id int) (*ent
...
+		if err != nil {
+			return retry.NonRetryableError(err)
+		}
+
+		archesSet := set.From(arches)
+		archesFound := set.New[string](0)
+
+		for _, resource := range allResources {
+			if resource.Name == fmt.Sprintf("%s/%s", os, release) {
+				archesFound.Insert(strings.Split(resource.Architecture, "/")[0])
```
tobiasdemendonca, line 260: Is it worth commenting in a couple of points here why exactly we're doing what we're doing? I don't follow 100% here but that could just be my lack of familiarity with the content.

tobiasdemendonca (changes requested):
Some minor comments.

tobiasdemendonca (approved):
Nice! Looks good

maas/resource_maas_boot_source_selection.go
```diff
@@ -59,15 +64,18 @@ func resourceMAASBootSourceSelection() *schema.Resource {
...
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Computed:    true,
+				Description: "The list of subarches for this selection. Default is: `[\"*\"]`",
 			},
 		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Minute),
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
```
SK1Y101, line 77: we may want to revisit these timeouts and make them user definable in future, for now i think 30 min default is fair though

maas/resource_maas_boot_source_selection.go
```diff
@@ -190,3 +240,87 @@ func getBootSourceSelection(client *client.Client, bootSource int, id int) (*ent
...
+					if !resourceSet.Complete {
+						return retry.RetryableError(fmt.Errorf("image still importing, waiting... "))
+					}
+				}
+			}
+		}
+
+		// There is still difference between user selected architectures set and unique boot resource architectures for the given
+		// os/release. The import is still ongoing.
+		if !archesSet.Equal(archesFound) {
```
SK1Y101, line 290: very much appreciate all the comments above, this is a non-trivial bit of logic to follow

maas/resource_maas_boot_source_selection_test.go
```diff
@@ -40,9 +39,26 @@ func TestAccResourceMAASBootSourceSelection_basic(t *testing.T) {
...
+				),
+			},
+			// Test update
+			{
+				Config: testAccMAASBootSourceSelection(os, release, updatedArches),
+				Check: resource.ComposeAggregateTestCheckFunc(append(
+					checks,
+					resource.TestCheckResourceAttr("maas_boot_source_selection.test", "arches.#", fmt.Sprintf("%v", len(updatedArches))),
+					resource.TestCheckResourceAttr("maas_boot_source_selection.test", "arches.0", updatedArches[0]),
+					resource.TestCheckResourceAttr("maas_boot_source_selection.test", "arches.1", updatedArches[1]),
```
SK1Y101, line 59: I'm assuming we can guarantee the ordering here?

SK1Y101 (commented):
changes look good to me, small questions just to clarify

maas/resource_maas_boot_source_selection.go
skatsaounis: By putting them in the schema, they are user definable. The 30 minutes here means: "If the user didn't provide any value, default to 30 minutes"

SK1Y101: Oh! my mistake, thanks!

maas/resource_maas_boot_source_selection_test.go
skatsaounis: That's a great catch. I updated the tests to tolerate any change in ordering.

tobiasdemendonca (approved):
Good catch

SK1Y101 (commented):
new changes look great, thanks for the response

PR #280, opened 2025-03-04
fix: 195 dns record removal

tobiasdemendonca:
## Description of changes

Fixes the issue of accidental deletion of both DNS resources that share a common IP address when one is deleted. This also fixes an error caused by a race condition when destroying both resources. 

Resolves https://github.com/canonical/terraform-provider-maas/issues/195
Supersedes https://github.com/canonical/terraform-provider-maas/pull/196

## Checklist

- [ ] I have written a PR title that follows the advice in DEVELOPMENT.md with the title format `type(scope): title`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module terraform-provider-maas
 
-go 1.22.7
+go 1.23
+
+toolchain go1.23.6
 
 require (
 	github.com/bflad/tfproviderlint v0.31.0
@@ -40,6 +42,7 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1 // indirect
 	github.com/hashicorp/go-plugin v1.6.2 // indirect
 	github.com/hashicorp/go-retryablehttp v0.7.7 // indirect
+	github.com/hashicorp/go-set/v3 v3.0.0 // indirect
 	github.com/hashicorp/go-uuid v1.0.3 // indirect
 	github.com/hashicorp/go-version v1.7.0 // indirect
 	github.com/hashicorp/hc-install v0.9.1 // indirect
--- a/maas/resource_maas_dns_record.go
+++ b/maas/resource_maas_dns_record.go
@@ -185,18 +185,18 @@ func resourceDnsRecordDelete(ctx context.Context, d *schema.ResourceData, meta i
 		return diag.FromErr(err)
 	}
 	if d.Get("type").(string) == "A/AAAA" {
-		dnsResource, err := client.DNSResource.Get(id)
-		if err != nil {
-			return diag.FromErr(err)
-		}
+		// dnsResource, err := client.DNSResource.Get(id)
+		// if err != nil {
+		// 	return diag.FromErr(err)
+		// }
 		if err := client.DNSResource.Delete(id); err != nil {
 			return diag.FromErr(err)
-		}
-		for _, ipAddress := range dnsResource.IPAddresses {
-			if err := client.IPAddresses.Release(&entity.IPAddressesParams{IP: ipAddress.IP.String()}); err != nil {
-				return diag.FromErr(err)
-			}
-		}
+		}	
+		// for _, ipAddress := range dnsResource.IPAddresses {
+		// 	if err := client.IPAddresses.Release(&entity.IPAddressesParams{IP: ipAddress.IP.String()}); err != nil {
+		// 		return diag.FromErr(err)
+		// 	}
+		// }
 	} else {
 		if err := client.DNSResourceRecord.Delete(id); err != nil {
 			return diag.FromErr(err)
--- a/maas/resource_maas_dns_record.go
+++ b/maas/resource_maas_dns_record.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/canonical/gomaasclient/client"
 	"github.com/canonical/gomaasclient/entity"
+	"github.com/hashicorp/go-set/v3"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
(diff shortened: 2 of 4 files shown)


maas/resource_maas_dns_record.go
```diff
@@ -206,6 +205,36 @@ func resourceDnsRecordDelete(ctx context.Context, d *schema.ResourceData, meta i
...
+		}
+	}
+	for _, ipAddress := range dnsResource.IPAddresses {
+		if allOtherDNSResourcesIPAddresses.Contains(ipAddress.IP.String()) {
+			continue
+		}
+		// Release the IP address if it is not used by another DNS resource.
+		// Terraform removes resources in parallel, so if it's already been deleted
+		// by another resource pointing to the same IP being removed, ignore it.
+		if err := client.IPAddresses.Release(&entity.IPAddressesParams{IP: ipAddress.IP.String()}); err != nil && !strings.Contains(err.Error(), "does not exist") {
```
tobiasdemendonca, line 231: The last part of this I found is required, as it stops the flaky errors produced when you run `terraform destroy` sometimes. 

```
err != nil && !strings.Contains(err.Error(), "does not exist")
```

This is because when terraform destroys resources, it does so concurrently, so sometimes I **think** it will remove the ip address of one whilst the other is being destroyed. This race condition causes an error, to reproduce, replace this part with 

```
err != nil
```

tobiasdemendonca, line 231: The last part of this I found is required, as it stops the flaky errors produced when you run `terraform destroy` sometimes. 

```
err != nil && !strings.Contains(err.Error(), "does not exist")
```

This is because when terraform destroys resources, it does so concurrently, so sometimes I **think** it will remove the ip address of one whilst the other is being destroyed. This race condition causes an error, to reproduce, replace this part with 

```
err != nil
```

maas/resource_maas_dns_record_test.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestAccResourceMAASDNSRecord_basic(t *testing.T) {
+	var dnsRecord entity.DNSResource
+	recordName := acctest.RandomWithPrefix("tf-dns-record-")
+	const recordName1 = "maas_dns_record.test"
+	const TEST_IP_ADDRESS = "8.8.8.8"
```
tobiasdemendonca, line 23: I've prioritised parallel tests here, and so had to use two separate test ips for each test, `TEST_IP_ADDRESS` and `TEST_IP_ADDRESS_2`. Without this, in my check destroy function, the tests would fail as one was being destroyed, but as the second test was being created, so the ip address was marked as allocated. Having two separate IPs keeps this from happening.

maas/resource_maas_dns_record_test.go
```diff
@@ -0,0 +1,195 @@
...
+			return fmt.Errorf("ip address is not released: %s with error: %s", ipAddress, err)
+		}
+		return nil
+	}
+}
+
+// Check if a particular IP address is allocated in MAAS
+func isIPAddressAllocated(conn *client.Client, ipAddress string) (bool, error) {
+	params := &entity.IPAddressesParams{IP: ipAddress}
+	allIPAddresses, err := conn.IPAddresses.Get(params)
```
skatsaounis, line 179: This is not the all IPs since you are using params to filter on the specific IP.

skatsaounis: If the IP does not exist anymore, you will not receive an error. Instead you will get back an empty slice. This is captured because the for loop is skipped and the function returns false, nil. In case of an error, then you should error out, since it is unexpected. As for the loop, if IP is still allocated, then we should expect that the slice has size 1 and the IP.String() matches the IP in question. In any other case, we also have to error out since this is unexpected.

maas/resource_maas_dns_record_test.go
skatsaounis: Let's rename the variables to camel case. In addition for the other test, `_2` suffix can be omitted. Parallel tests make sense here and indeed, using two different IP addresses is the appropriate way to go.

maas/resource_maas_dns_record.go
skatsaounis: That's fine. I am resolving this.

go.mod
```diff
@@ -1,6 +1,8 @@
 module terraform-provider-maas
 
-go 1.22.7
+go 1.23
```
skatsaounis, line 3: If we are going to use go-set/v3, we also need to bump to Go 1.23 without including the toolchain.

maas/resource_maas_dns_record.go
```diff
@@ -206,6 +205,36 @@ func resourceDnsRecordDelete(ctx context.Context, d *schema.ResourceData, meta i
 	return nil
 }
 
+// Release all IP addresses of a DNS resource, but only if it is not used by other DNS resources.
+// NOTE: exactly why this is required here and not in MAAS itself is not clear.
```
skatsaounis, line 209: This message should be removed.

go.mod
```diff
@@ -71,6 +72,7 @@ require (
 	github.com/oklog/run v1.0.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/posener/complete v1.2.3 // indirect
+	github.com/shoenig/test v1.11.0 // indirect
```
tobiasdemendonca, line 75: Dependency of go-set v2

tobiasdemendonca:
@skatsaounis this is ready for review

maas/resource_maas_dns_record_test.go
```diff
@@ -0,0 +1,201 @@
...
+	})
+}
+
+func getDNSRecordConfigSameIPA_AAAA(resourceName1 string, resourceName2 string, recordName1 string, recordName2 string, ipAddress string) string {
+	return fmt.Sprintf(`
+	resource "maas_dns_record" %q {
+	  name   = %q
+	  type   = "A/AAAA"
+	  data   = %q
+	  domain = "maas"
```
skatsaounis, line 87: let's create a random domain as part of the tests since there is no guarantee that MAAS under test is using it. The change is trivial and should make the CI happy :slightly_smiling_face: Same thing applies to all places where `maas` domain is considered as something that preexists.

skatsaounis (changes requested):

skatsaounis (approved):
LGTM +1

PR #281, opened 2025-03-06
fix: 195 dns record removal revised

tobiasdemendonca:
## Description of changes

*A clear and concise description of your changes here.*

*A link to an issue or ticket that this PR is addressing.*

## Checklist

- [ ] I have written a PR title that follows the advice in DEVELOPMENT.md with the title format `type(scope): title`

### DIFF
--- a/maas/resource_maas_device.go
+++ b/maas/resource_maas_device.go
@@ -114,6 +114,7 @@ func resourceDeviceCreate(ctx context.Context, d *schema.ResourceData, meta inte
 		Description:  d.Get("description").(string),
 		Domain:       d.Get("domain").(string),
 		Hostname:     d.Get("hostname").(string),
+		MacAddresses: expandNetworkInterfacesItems(d.Get("network_interfaces").(*schema.Set).List()),
 	}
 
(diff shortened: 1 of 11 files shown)


maas/resource_maas_device.go
```diff
@@ -134,6 +134,7 @@ func resourceDeviceUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 		Domain:      d.Get("domain").(string),
 		Hostname:    d.Get("hostname").(string),
 		Zone:        d.Get("zone").(string),
+		// NetworkInterfaces: ...
```
tobiasdemendonca, line 137: @skatsaounis in the standup you suggested this function would result in a bug, I think because you thought we'd be double counting network interfaces if another resource network_interface_phsysical was defined, for this particular device. However, we aren't actually tracking network interfaces here. This is a similar bug to what can be seen in https://github.com/canonical/terraform-provider-maas/issues/273 . 

Do you think we should aim to fix this issue now? It occurs in a few places so I would presume a similar design choice could apply to all

PR #290, opened 2025-03-20
feat: 230 extend tagging to support nics

tobiasdemendonca:
## Description of changes

An additional resource to manage the tags of network interfaces that aren't managed with terraform. This resource must not be used in conjunction with the nested tag keys in resources such as `maas_network_interface_physical`. Other examples of additional supporting resources that follow this design pattern can be found in the [AWS provider](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/resource-tagging#ignoring-changes-to-specific-tags).

Resolves part of https://github.com/canonical/terraform-provider-maas/issues/230 

## Checklist


docs/resources/network_interface_tag.md
```diff
@@ -0,0 +1,30 @@
...
+
+
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `interface_id` (Number) The network interface ID to tag.
+- `tags` (Set of String) The tags to assign to the network interface. It should be short and without spaces.
```
tobiasdemendonca, line 21:
```suggestion
- `tags` (Set of String) The tags to assign to the network interface. Tag names should be short and without spaces.
```

maas/resource_maas_network_interface_tag.go
```diff
@@ -0,0 +1,204 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceMaasNetworkInterfaceTag() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Provides a resource to manage a MAAS tags as strings on a network interface not managed by Terraform. It is highly recommended to not use this resource to manage tags on network interfaces managed by Terraform, as this will cause conflicts and will overwrite the tags already set. Use the nested `tags` attribute on resources such as `maas_network_interface_physical` if you need to manage tags on network interfaces managed by Terraform.",
```
tobiasdemendonca, line 16:
```suggestion
		Description:   "Provides a resource to manage tags as strings on a network interface that is not managed by Terraform. It is highly recommended to not use this resource to manage tags on network interfaces that are already managed by Terraform, as this will cause conflicts and will overwrite the tags already set. Use the nested `tags` attribute on resources such as `maas_network_interface_physical` if you need to to do this.
```

tobiasdemendonca:
Note: example required

maas/resource_maas_network_interface_tag.go
```diff
@@ -0,0 +1,204 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceMaasNetworkInterfaceTag() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Provides a resource to manage tags as strings on a network interface that is not managed by Terraform. It is highly recommended to not use this resource to manage tags on network interfaces that are already managed by Terraform, as this will cause conflicts and will overwrite the tags already set. Use the nested `tags` attribute on resources such as `maas_network_interface_physical` if you need to to do this.",
```
skatsaounis, line 16: What about changing the description in a way that: 1) we allow the usage of tags for interfaces created by Terraform. What is stopping us to chain 2 terraform modules where the first one is adding the interfaces and the second one is deploying the machines and is applying tags afterwards? We do not have such limitation. On the other hand 2) If users want to create interfaces with tags and later on manage tags with the separate resource, they have to add lifecycle ignore changes:

```
resource "maas_network_interface_physical" "example" {
  # ...
  tags = ["one", "two"]
  lifecycle {
    ignore_changes = [
      # Ignore changes to tags, e.g. because we are updating them later on with the dedicated resource
      tags,
    ]
  }
}
```

In addition, this is something that can be introduced with a template rather than injected in Go. Please consult: https://github.com/hashicorp/terraform-plugin-docs for conventional paths, template fields, etc.

examples/resources/maas_network_interface_tag/resource.tf
```diff
@@ -0,0 +1,27 @@
+resource "maas_network_interface_tag" "test" {
+  machine      = "abc123"
+  interface_id = "12"
```
skatsaounis, line 3: based on your implementation this is a number. What about removing the double quotas from every `interface_id`?

maas/resource_maas_network_interface_tag.go
```diff
@@ -0,0 +1,204 @@
...
+			return diag.FromErr(err)
+		}
+	}
+	d.SetId("")
+	return nil
+}
+
+// Split the state ID of a tag in the format system_id:interface_id into its component ids, where system_id is the system ID of the machine or device, and interface_id is the ID of the network interface.
+func SplitTagStateId(stateId string) (string, int, error) {
+	splitId := strings.SplitN(stateId, ":", 2)
```
skatsaounis, line 195: we expect user to provide `/` separator in the example but here you are splitting based on `:`. Let's make sure that we use `/` everywhere, including docs, examples, implementation and tests. Maybe consider it too for `d.SetId`

maas/resource_maas_network_interface_tag.go
```diff
@@ -0,0 +1,204 @@
...
+	if existingInterface.Tags == nil {
+		existingInterface.Tags = []string{}
+	}
+	for _, tag := range existingInterface.Tags {
+		if !slices.Contains(desiredTags, tag) {
+			client.NetworkInterface.RemoveTag(systemId, interfaceId, tag)
+		}
+	}
+
+	// Add tags that are in the desired set. AddTag will not add duplicates.
```
skatsaounis, line 111: But still you know which tags you can skip adding. What about adding a similar check to remove in order to avoid unnecessary HTTP calls to MAAS?

maas/resource_maas_network_interface_tag.go
```diff
@@ -0,0 +1,204 @@
...
+	desiredTags := convertToStringSlice(d.Get("tags").(*schema.Set).List())
+
+	// Remove tags that are not in the specified set
+	for _, tag := range existingTags {
+		if !slices.Contains(desiredTags, tag) {
+			client.NetworkInterface.RemoveTag(systemId, interfaceId, tag)
+		}
+	}
+
+	// Add tags that are in the specified set. AddTag will not add duplicates.
```
skatsaounis, line 168: Same comment as create.

maas/resource_maas_network_interface_tag.go
```diff
@@ -0,0 +1,204 @@
...
+func resourceNetworkInterfaceTagDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*ClientConfig).Client
+	systemId, err := getMachineOrDeviceSystemID(client, d)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	interfaceId := d.Get("interface_id").(int)
+	tags := convertToStringSlice(d.Get("tags").(*schema.Set).List())
+	for _, t := range tags {
+		_, err := client.NetworkInterface.RemoveTag(systemId, interfaceId, t)
```
skatsaounis, line 184: what is the MAAS reaction if the tag tried to be removed is not applied to the machine or device the moment we are trying to remove it? (deleted outside of Terraform prior to `terraform destroy`)

maas/resource_maas_network_interface_tag_test.go
```diff
@@ -0,0 +1,138 @@
...
+	if systemId != expectedSystemId || interfaceId != expectedInterfaceId {
+		t.Fatalf("Expected system ID %s and interface ID %d, got system ID %s and interface ID %d", expectedSystemId, expectedInterfaceId, systemId, interfaceId)
+	}
+}
+
+func TestAccNetworkInterfaceTag_basic(t *testing.T) {
+	hostname := acctest.RandomWithPrefix("tf")
+	macAddress := testutils.RandomMAC()
+	tagName := acctest.RandomWithPrefix("tag")
+	tagName2 := acctest.RandomWithPrefix("tag2")
```
skatsaounis, line 32: tag2 can also be tag since it is a prefix. Let's remove 2 and 3 from the next line

maas/resource_maas_network_interface_tag_test.go
```diff
@@ -0,0 +1,138 @@
...
+			{
+				Config: testAccMaasNetworkInterfaceTagConfig(hostname, macAddress, tagName2, tagName3),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckMaasNetworkInterfaceTagExists("maas_network_interface_tag.test"),
+					resource.TestCheckResourceAttr("maas_network_interface_tag.test", "tags.#", "2"),
+					resource.TestCheckTypeSetElemAttr("maas_network_interface_tag.test", "tags.*", tagName2),
+					resource.TestCheckTypeSetElemAttr("maas_network_interface_tag.test", "tags.*", tagName3),
+				),
+			},
+			// Test import.
```
skatsaounis, line 60: How is this test going to work? I would expect we specify to Terraform how to calculate the ID with `ImportStateIdFunc` like we do in couple other tests

maas/resource_maas_network_interface_tag_test.go
```diff
@@ -0,0 +1,138 @@
...
+	return fmt.Sprintf(`
+resource "maas_device" "test" {
+  hostname = %q
+  network_interfaces {
+	mac_address = %q 
+  }
+}
+
+resource "maas_network_interface_tag" "test" {
+  device = maas_device.test.id
```
skatsaounis, line 80: this is not proper tf formatted

maas/resource_maas_network_interface_tag_test.go
```diff
@@ -0,0 +1,138 @@
...
+  interface_id = [for iface in maas_device.test.network_interfaces : iface.id if iface.mac_address == %q][0]
+  tags = %s
+}
+	`, hostname, macAddress, macAddress, fmt.Sprintf("[\"%s\"]", strings.Join(tagNames, "\", \"")))
+}
+
+func testAccCheckMaasNetworkInterfaceTagExists(resourceName string) resource.TestCheckFunc {
```
skatsaounis, line 87: were is the test for the tag? I have a feeling that this function is incomplete

skatsaounis (changes requested):

maas/resource_maas_device.go
```diff
@@ -27,7 +27,6 @@ func resourceMaasDevice() *schema.Resource {
 				return []*schema.ResourceData{d}, nil
 			},
 		},
-
```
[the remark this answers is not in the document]
tobiasdemendonca: I ran go fmt on everything so this good be the reason

maas/resource_maas_network_interface_tag.go
tobiasdemendonca: Sorry this is a clear oversight from me just copying from a similar file `maas_network_interface_physical/import.sh`, which by chance is one of the only files where we have this format and the rest are `:`. Do you mind me sticking to the what is the main pattern and keeping the colon over `/`?

maas/resource_maas_network_interface_tag.go
tobiasdemendonca: `client.NetworkInterface.RemoveTag` handles this without an error :+1:

maas/resource_maas_network_interface_tag.go
tobiasdemendonca: Added a new guides section as this spans multiple resources

maas/resource_maas_network_interface_tag_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"terraform-provider-maas/maas"
+	"terraform-provider-maas/maas/testutils"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestSplitTagStateId(t *testing.T) {
+	expectedSystemId := "acb123"
```
SK1Y101, line 18: I know it doesn't actually matter but i would like this to be `abc123` (:

maas/resource_maas_network_interface_tag_test.go
```diff
@@ -0,0 +1,161 @@
...
+resource "maas_device" "test" {
+  hostname = %q
+  network_interfaces {
+    mac_address = %q
+  }
+}
+
+resource "maas_network_interface_tag" "test" {
+  device       = maas_device.test.id
+  interface_id = [for iface in maas_device.test.network_interfaces : iface.id if iface.mac_address == %q][0]
```
SK1Y101, line 94: nice use of list comprehension!

maas/utils.go
```diff
@@ -141,3 +141,27 @@ func getMachineOrDeviceSystemID(client *client.Client, d *schema.ResourceData) (
...
+
+// Get the type of the relevant entity from the system ID, by checking if the device or machine exists in MAAS.
+// This gets all devices, then all machines, so is not efficient and should be used sparingly.
+func getMachineOrDeviceTypeFromSystemID(client *client.Client, systemID string) (string, error) {
+	device, err := getDevice(client, systemID)
+	if err == nil && device != nil {
+		return "device", nil
+	}
+
+	if !strings.Contains(err.Error(), "404 Not Found") {
```
SK1Y101, line 153: what is preventing this triggering if the systemID was for a machine?

maas/resource_maas_network_interface_tag_test.go
```diff
@@ -0,0 +1,161 @@
...
+		// Check the interface doesn't exist
+		response, err := conn.NetworkInterface.Get(systemId, interfaceId)
+		if err == nil {
+			if response != nil && response.ID == interfaceId {
+				return fmt.Errorf("MAAS Network Interface (%s) still exists.", rs.Primary.ID)
+			}
+		}
+		// If the error is a 404, the interface is destroyed as expected
+		if !strings.Contains(err.Error(), "404 Not Found") {
+			return err
```
SK1Y101, line 157: I was about to ask if this should be returning the error if this is expected behaviour before I noticed the `!`. Should the comment be changed slightly so it matches the same logical flow as the program?

ie: `If the error is *not* a 404, the interface has not been destroyed as expected`

SK1Y101 (approved):
As best I can tell this all looks good! nice!

minor nit about comment logic vs program logic, but does not affect the code at all

maas/utils.go
tobiasdemendonca: Very very good catch

tobiasdemendonca: It's not a 404 Not Found, it's "device (id) not found"

tobiasdemendonca: If the system id was for a machine, it should be (now the correct) device not found error, it should continue on. This works with both machines and devices

SK1Y101: Nope, just the control flow of:
machine id -> device get -> 404 not found -> error thrown and halted prematurely

tobiasdemendonca: Machine id -> device get -> not found error -> continue, as error contains "not found" and this if check only executes if the error is not a not found error

SK1Y101: okay, cool!

maas/resource_maas_network_interface_tag.go
skatsaounis: I reopened this one. If we want to stick with a single one, that is `/`. The `:` is problematic for any import that can accept a MAC address. I fixed it for physical interfaces back then but other interface resources are still experiencing it https://github.com/canonical/terraform-provider-maas/pull/169

skatsaounis (approved):
LGTM +1

Issue #215, opened 2024-07-06
Missing LVM configuration support


aym-frikha:
Using this terraform provider , there is no way to configure Volume groups and LVM devices on enlisted and commissioned machines. 

The idea is to create LVM partitions for /tmp, /var ... on raid1 devices.

skatsaounis:
Hi @aym-frikha. Thank you for the suggestion. In order to enable the support for LVM, gomaasclient must support first Volume groups.

skatsaounis:
In addition to that, `set_storage_layout` should also be added to machine entity at `gomaasclient`

PR #293, opened 2025-03-27
Feat: Introduce volume group resource

SK1Y101:
## Description of changes

Add resource for maas_volume_groups, and their associated CRUD operations

addresses part of #215, as logical volumes require volume groups to build on top of.

## Checklist


maas/utils.go
```diff
@@ -39,6 +39,16 @@ func convertToStringSlice(field interface{}) []string {
...
+func convertSliceToStringSlice(field []interface{}) []string {
+	if field == nil {
+		return nil
+	}
+	result := make([]string, len(field))
+	for i, value := range field {
+		result[i] = value.(string)
+	}
+	return result
+}
```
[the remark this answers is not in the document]
SK1Y101: That's probably a good idea, yeah!

maas/resource_maas_volume_group.go
```diff
@@ -0,0 +1,205 @@
...
+		return diag.FromErr(err)
+	}
+	id, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	var addBlockDevices []string
+	var removeBlockDevices []string
+	if d.HasChange("block_Devices") {
```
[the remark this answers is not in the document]
SK1Y101: I noticed this too, you would not believe how long I spent tearing my hair out wondering why no differences were detected ):

maas/resource_maas_volume_group.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+func resourceMaasVolumeGroup() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Provides a resource to manage MAAS Volume Groups, and construct them from partion-less block devices.",
+		CreateContext: resourceMaasVolumeGroupCreate,
+		ReadContext:   resourceMaasVolumeGroupRead,
+		UpdateContext: resourceMaasVolumeGroupUpdate,
+		DeleteContext: resourceMaasVolumeGroupDelete,
+
```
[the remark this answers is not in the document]
SK1Y101: Uncertain. I think I'll add it, and we can remove it in future if it's not needed, rather than the inverse

maas/resource_maas_volume_group.go
```diff
@@ -61,6 +66,44 @@ func resourceMaasVolumeGroup() *schema.Resource {
 	}
 }
 
+func resourceMaasVolumeGroupImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	idParts := strings.Split(d.Id(), ":")
+
+	if len(idParts) != 2 || idParts[0] == "" || idParts[1] == "" {
+		return nil, fmt.Errorf("unexpected format of ID (%q), expected MACHINE_ID:VOLUME_GROUP_ID", d.Id())
+	}
```
tobiasdemendonca, line 74: If you were using `/` as a delimiter we could just rename `SplitTagStateId` and use this function. However given we've got a few places to change this so we will come back to change this in the future for a number of resources, and CI is being finicky, I think this is okay

maas/resource_maas_volume_group.go
```diff
@@ -0,0 +1,271 @@
...
+		},
+	}
+}
+
+func resourceMaasVolumeGroupImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	idParts := strings.Split(d.Id(), ":")
+
+	if len(idParts) != 2 || idParts[0] == "" || idParts[1] == "" {
+		return nil, fmt.Errorf("unexpected format of ID (%q), expected MACHINE_ID:VOLUME_GROUP_ID", d.Id())
+	}
```
tobiasdemendonca, line 74: If you were using `/` as a delimiter we could just rename `SplitTagStateId` and use this function. However given we've got a few places to change this import id format, we will come back to change this in the future for a number of resources, and CI is being finicky, I think this is okay for now

examples/resources/maas_volume_group/import.sh
```diff
@@ -0,0 +1,2 @@
+# Volume groups can be imported with the machine identifier (system ID, hostname, or FQDN) and the volume group identifier (ID or name). e.g.
+$ terraform import maas_block_device.vdb machine-06:vg1
```
tobiasdemendonca, line 2: See later comment regarding import id use

maas/resource_maas_volume_group.go
tobiasdemendonca: Note, Stamatis said we prefer /'s over : because `/` doesn't compete with mac addresses, see [this comment](https://github.com/canonical/terraform-provider-maas/pull/290#discussion_r2013938370)

maas/resource_maas_volume_group.go
```diff
@@ -0,0 +1,271 @@
...
+		"machine":        volumeGroup.SystemID,
+		"name":           volumeGroup.Name,
+		"size":           volumeGroup.Size,
+		"used_size":      volumeGroup.UsedSize,
+		"available_size": volumeGroup.AvailableSize,
+		"uuid":           volumeGroup.UUID,
+	}
+	if err := setTerraformState(d, tfState); err != nil {
+		return nil, err
+	}
```
tobiasdemendonca, line 100: This is more of a comment rather than a suggestion, as I think we're doing this throughout, but I believe the role of the import is just to import the id, and then let read handle the rest (see [here](https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-resource-import#implement-import-functionality). Therefore, this isn't required. However, this does mean that read shouldn't have any `d.get`s, which would require a bit of a change. I think this is okay

maas/resource_maas_volume_group_test.go
```diff
@@ -0,0 +1,169 @@
...
+			},
+		},
+	})
+}
+
+func testAccMaasVolumeGroup(machine string, name string, blockDevices []string) string {
+	return fmt.Sprintf(`
+
+data "maas_machine" "machine" {
+  hostname = "%s"
```
tobiasdemendonca, line 79: really minor, you can use `%q` instead of `"%s"`, not essential to do

tobiasdemendonca (approved):
Some minor comments that I don't think are essential to change, looks good

maas/resource_maas_volume_group_test.go
SK1Y101: No, I like that, reduces redundancy

maas/resource_maas_volume_group.go
SK1Y101: I think it's a fair point, we have a dependency on the machine ID, but nothing else, I should leave it to the read.

maas/resource_maas_volume_group.go
SK1Y101: I think it's probably a good idea to unify across stuff, I'll change it

examples/resources/maas_volume_group/import.sh
```diff
@@ -0,0 +1,2 @@
+# Volume groups can be imported with the machine identifier (system ID, hostname, or FQDN) and the volume group identifier (ID or name). e.g.
+$ terraform import maas_block_device.vdb machine-06/vg1
```
skatsaounis, line 2: this is an example of a block_device and as such it needs to change

maas/resource_maas_volume_group.go
```diff
@@ -0,0 +1,259 @@
...
+		CreateContext: resourceMaasVolumeGroupCreate,
+		ReadContext:   resourceMaasVolumeGroupRead,
+		UpdateContext: resourceMaasVolumeGroupUpdate,
+		DeleteContext: resourceMaasVolumeGroupDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceMaasVolumeGroupImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"available_size": {
```
skatsaounis, line 28: Available size looks like something dynamic. Maybe it is better to only keep `size`?

maas/resource_maas_volume_group.go
```diff
@@ -0,0 +1,259 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The name for this volume group",
+			},
+			"size": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "The volume group size (B).",
+			},
+			"used_size": {
```
skatsaounis, line 55: Used size looks like something dynamic. Maybe it is better to only keep `size`?

maas/resource_maas_volume_group.go
```diff
@@ -0,0 +1,259 @@
...
+				Required:    true,
+				ForceNew:    true,
+				Description: "The machine identifier (system ID, hostname, or FQDN) that owns the volume group.",
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The name for this volume group",
+			},
+			"size": {
```
skatsaounis, line 50: Maybe we need to convert it in GB like we do for maas_block_device: `Size:      int64(d.Get("size_gigabytes").(int)) * 1024 * 1024 * 1024`. Is there any benefit of keeping it in bytes?

skatsaounis: Or even better, what if we consume `HumanSize` from the API response?

maas/resource_maas_volume_group_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testutils.PreCheck(t, []string{"TF_ACC_BLOCK_DEVICE_MACHINE"}) },
+		Providers:    testutils.TestAccProviders,
+		CheckDestroy: testAccCheckMaasVolumeGroupDestroy,
+		ErrorCheck:   func(err error) error { return err },
+		Steps: []resource.TestStep{
+			// Test initial creation
+			{
+				Config: testAccMaasVolumeGroup(machine, name, []string{"maas_block_device.bd1.id"}),
```
skatsaounis, line 37: we are missing a check for the size on the initial creation and the update

maas/resource_maas_volume_group.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func findVolumeGroupBlockDevices(volumeGroup *entity.VolumeGroup) []string {
+	var blockDevices []string
+	for _, device := range volumeGroup.Devices.([]interface{}) {
+		thisDevice := device.(map[string]interface{})
+
+		var deviceId string
+
+		// partitions list a device id of the parent block device
```
skatsaounis, line 231: This is not clear to me. So we can create logical groups which contain partitions by providing the id of the partition and in the read response, instead of that id, we get the id of the parent block device? Maybe I miss-understand it. Could you please provide an example with some JSON responses?

skatsaounis (changes requested):
I have added a question and some suggestions for improvement. Other than that, we are almost good to go :+1:

skatsaounis (changes requested):
It seems that MAAS does create a partition to the boot disk of a machine, if the boot disk is specified as a volume group block device. This should be avoided by Terraform since it has unpredictable behavior. We should instead add a validation and error out if the Terraform practitioner set it.

In addition, we should allow the volume group to accept both block devices and partitions and expect at least one to be set to at least one block_device or partition.

UPDATE: We can check if the machine boot_disk is included in the block_devices that the Terraform practitioner has supplied

tobiasdemendonca (dismissed):
LGTM

maas/resource_maas_volume_group_test.go
```diff
@@ -0,0 +1,190 @@
...
+  id_path        = "/dev/bd1"
+  is_boot_device = true
+
+  partitions {
+    size_gigabytes = 20
+  }
+
+  lifecycle {
+    ignore_changes = [
+      partitions[0].fs_type
```
skatsaounis, line 99: Q: Why do we need this?

maas/resource_maas_volume_group_test.go
```diff
@@ -0,0 +1,190 @@
...
+  id_path        = "/dev/bd2"
+}
+
+resource "maas_volume_group" "test" {
+  machine       = data.maas_machine.machine.id
+  name          = %q
+  block_devices = [%s]
+  partitions 	= [%s]
+
+  depends_on = [maas_block_device.bd1, maas_block_device.bd2]
```
skatsaounis, line 118: Q: Why do we need this?

maas/resource_maas_volume_group_test.go
SK1Y101: MAAS automatically creates a filesystem on the boot partition, terraform doesn't like that.

You also can't supply a partition with a filesystem for MAAS to create a volume group, so we can't pre-populate this with the correct value :(

maas/resource_maas_volume_group_test.go
SK1Y101: I don't know that this is required anymore, I'll test without it

maas/resource_maas_volume_group_test.go
skatsaounis: Interesting. Would you mind point me to the MAAS logic on the automatic creation of the filesystem?

SK1Y101: The filesystem type is defined in https://github.com/canonical/maas/blob/master/src/maasserver/enum.py#L576
but i can't find where it's further used

examples/resources/maas_volume_group/resource.tf
```diff
@@ -0,0 +1,36 @@
...
+  size_gigabytes = 35
+  tags = [
+    "ssd",
+  ]
+
+  partitions {
+    size_gigabytes = 30
+  }
+
+  lifecycle {
```
skatsaounis, line 24: This one should go away from both the example and the docs

skatsaounis (approved):
LGTM +1

PR #294, opened 2025-03-31
feat: Introduce LVM resource management

SK1Y101:
## Description of changes

Add resource for maas LVM storage layouts, and their associated CRUD operations:

```terraform
resource "maas_logical_volume" "test" {
  fs_type 		 = "etx4"
  machine 		 = data.maas_machine.machine.id
  mount_point            = "/var"
  mount_options          = "defaults,noatime"
  name 			 = "test LVM"
  size_gigabytes         = 70
  volume_group 	         = maas_volume_group.test.id
}
```

will create an LVM on top of volume_group.test

Requires #293 to land first.

Resolves: #215, adding LVM support 

## Checklist


skatsaounis (changes requested):
The PR looks good so far. Let me compose a list of things that are missing:

- we need to support mounting the logical volume like we do with physical block devices
- we need to support update function
- what is the purpose of adding the suffix `lvm` everywhere? I think that we have to remove it as logical volume (LV) is a key component of LVM. If we had to introduce it, then prefix makes more sense, but at the same time we would have to add it as prefix to volume group (VG) as well.

SK1Y101:
>     * we need to support mounting the logical volume like we do with physical block devices
Should this be an operation, or a property like `mount_point` on the LVM

>     * we need to support update function
The API, and as far as i can tell MAAS itself, doesn't really have a notion of updating an LVM. it's treated like a blockdevice, so i assume we can use BD operations, but what would you actually update?

>     * what is the purpose of adding the suffix `lvm` everywhere? I think that we have to remove it as logical volume (LV) is a key component of LVM. If we had to introduce it, then prefix makes more sense, but at the same time we would have to add it as prefix to volume group (VG) as well.
Yeah fair, updated and removed

maas/resource_maas_logical_volume.go
```diff
@@ -0,0 +1,215 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"fs_type": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The file system type (e.g. `ext4`). If this is not set, the volume is unformatted.",
+			},
+			"machine": {
+				Type:        schema.TypeString,
+				Required:    true,
```
tobiasdemendonca, line 31: Could be debatable but looking at `maas_block_device` this I would perhaps expect this to be the case if the idea of updating to a new machine doesn't make sense
```suggestion
				Required:    true,
				ForceNew:    true,
```

tobiasdemendonca: This seems to reflect in your update function I think

maas/resource_maas_logical_volume.go
```diff
@@ -0,0 +1,215 @@
...
+		return diag.FromErr(err)
+	}
+
+	id, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	params := entity.BlockDeviceParams{
+		Name: d.Get("name").(string),
```
tobiasdemendonca, line 135: Either you're missing "size" here or I think it should be `ForceNew`

maas/resource_maas_logical_volume_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestAccResourceMAASLogicalVolume_basic(t *testing.T) {
+	var LogicalVolume entity.BlockDevice
+
+	machine := os.Getenv("TF_ACC_BLOCK_DEVICE_MACHINE")
+
+	size := 70
```
tobiasdemendonca, line 22: Nit: Not familiar with how big these things normally are but should we be making this smaller? I know we've got a lot of space but this seems kinda wildly big imo

maas/resource_maas_logical_volume_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+func testAccMAASLogicalVolume(machine string, fsType string, name string, size int, mountPoint string) string {
+	return fmt.Sprintf(`
+data "maas_machine" "machine" {
+  hostname = %q
+}
+
+resource "maas_block_device" "lvm_bd1" {
+  machine        = data.maas_machine.machine.id
+  name           = "lvm_bd1"
```
tobiasdemendonca, line 73: if we can have block devices of the same name this is fine otherwise i'd suggest adding a random prefix

maas/resource_maas_logical_volume_test.go
```diff
@@ -0,0 +1,176 @@
...
+  size_gigabytes = 25
```
tobiasdemendonca, line 74: Similar suggestion for this whole config on if we need to allocate this amount of space for this test

maas/resource_maas_logical_volume_test.go
tobiasdemendonca: Same for other names

maas/resource_maas_logical_volume_test.go
```diff
@@ -0,0 +1,176 @@
...
+		if !ok {
+			return fmt.Errorf("Could not find machine id on resource")
+		}
+
+		gotLogicalVolume, err := conn.BlockDevice.Get(machine, id)
+		if err != nil {
+			return fmt.Errorf("error getting the logical volume: %s", err)
+		}
+
+		*logicalVolume = *gotLogicalVolume
```
tobiasdemendonca, line 138: Do we need this at all, I've seen this done in other tests but I don't see the point if it's not being used later, even defining this variable at the start of the tests
```suggestion
```

tobiasdemendonca (changes requested):
Your PR reads super nicely, I'm not that familiar with LVMs but can get the idea from just the clear function names with few comments, lovely stuff. 

A few comments here and there

maas/resource_maas_logical_volume.go
SK1Y101: Hmm yeah, that's a good point

docs/resources/logical_volume.md
```diff
@@ -0,0 +1,74 @@
...
+
+# maas_logical_volume (Resource)
+
+
+
+## Example Usage
+
+```terraform
+resource "maas_block_device" "vdb1" {
+  machine        = maas_machine.virsh_vm2.id
```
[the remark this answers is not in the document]
SK1Y101: Yep, excellent point, will add that

maas/resource_maas_logical_volume_test.go
SK1Y101: Oh right yeah, I was just throwing numbers around. will decrease this by an order of magnitude

maas/resource_maas_logical_volume.go
SK1Y101: I don't think you can change the size of an LVM, maybe shrink it, but definitely can't grow it beyond the volume groups size.

skatsaounis:
Blocked by: https://github.com/canonical/gomaasclient/issues/123

maas/resource_maas_logical_volume_test.go
```diff
@@ -0,0 +1,172 @@
...
+  name          = %q
+  block_devices = [maas_block_device.lvm_bd2.id]
+  partitions 	= [maas_block_device.lvm_bd1.partitions.0.id]
+}
+
+resource "maas_logical_volume" "test" {
+  fs_type 		 = %q
+  machine 		 = data.maas_machine.machine.id
+  name 			 = %q
+  volume_group 	 = maas_volume_group.lvm_vg.id
```
skatsaounis, line 101: I think we need `terraform fmt` here

maas/resource_maas_logical_volume.go
```diff
@@ -0,0 +1,219 @@
...
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The name for this logical volume",
+			},
+			"size_gigabytes": {
+				Type:        schema.TypeInt,
+				Required:    true,
+				ForceNew:    true,
```
skatsaounis, line 53: This `ForceNew` should be removed. Most probably it is a remnant from previous commits when the gomaasclient in use could not properly set the size

skatsaounis (changes requested):
Some minor changes and we are good to go! Plus, please update to latest master

maas/resource_maas_logical_volume.go
```diff
@@ -0,0 +1,218 @@
...
+			"machine": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The machine identifier (system ID, hostname, or FQDN) that owns the volume group.",
+			},
+			"mount_options": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The options used for the logical volume mount.",
```
tobiasdemendonca, line 38: Just for my understanding is this comma separated mount options?

tobiasdemendonca (commented):
Minor nit

maas/resource_maas_logical_volume.go
```diff
@@ -0,0 +1,219 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The volume group identifier (ID or name) to apply this logical volume on top of.",
+			},
+		},
+	}
+}
+
+func resourceMAASLogicalVolumeCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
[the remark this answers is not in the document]
SK1Y101: Oh, fair, yeah they do! thanks

maas/resource_maas_logical_volume.go
SK1Y101: Yes, i believe so, I'll double check and then specify

tobiasdemendonca (approved):
LGTM

PR #298, opened 2025-04-01
feat: add `maas_devices` data source

ricardbejarano:
## Description of changes
Adds the `maas_devices` data source, which lists all devices visible to the user.

## Issue or ticket link (if applicable)
Resolves: #241 (partially, `maas_machines` is addressed in #299)

## Checklist

- [X] I have written a PR title that follows the advice in DEVELOPMENT.md with the title format `type(scope): title`
- [X] I have written a description of the changes in the PR that is clear and concise.
- [X] I have updated the documentation to reflect the changes.
- [X] I have added or updated the tests to reflect the changes.
- [X] I have run the unit tests and they pass.
- [X] I have run the acceptance tests and they pass.

Issue #241, opened 2024-10-29
Add `maas_devices`, `maas_machines` data sources


ricardbejarano:
This would map to the listing operations on both devices and machines, so that we can list what devices/machines are visible to the user.

ricardbejarano:
shut up bot

ricardbejarano:
shh!

ricardbejarano:
Anyone? @skatsaounis @troyanov @SK1Y101 maybe?

troyanov:
@ricardbejarano it is indeed getting a bit annoying. I will review our workflow and probably add exceptions for certain tags. Sorry about it.

PR #299, opened 2025-04-01
feat: add `maas_machines` data source

ricardbejarano:
## Description of changes
Adds the `maas_machines` data source, which lists all machines visible to the user.

#241 (partially, `maas_devices` is addressed in #298)

## Checklist

- [X] I have written a PR title that follows the advice in DEVELOPMENT.md with the title format `type(scope): title`
- [X] I have written a description of the changes in the PR that is clear and concise.
- [X] I have updated the documentation to reflect the changes.


ricardbejarano:
I haven't added tests because `dataSourceMaasMachine` doesn't have them either.
What should we do?

cc @skatsaounis @tobiasdemendonca (going by GitHub Insights)

SK1Y101:
@ricardbejarano I think it would be good to add a test file, so that we them for all of them moving forward

ricardbejarano:
There you go, lemme know if that's not sufficient.

ricardbejarano:
This is ready for review @tobiasdemendonca

maas/data_source_maas_machines.go
```diff
@@ -0,0 +1,97 @@
...
+							Computed:    true,
+							Description: "The power management type (e.g. `ipmi`) of the machine.",
+						},
+						"pxe_mac_address": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "The MAC address of the machine's PXE boot NIC.",
+						},
+						"zone": {
+							Type:        schema.TypeString,
```
tobiasdemendonca, line 58: so these would be reduced to just 

```go
"system_id": {
    Type: schema.TypeString,
    Computed: true,
    Description: "The system ID of the machine.",
}
"hostname": {
   Type: schema.TypeString,
   Computed: true,
   Description: "The machine hostname.",
}
```

tobiasdemendonca (changes requested):
Hi @ricardbejarano , thank you very much for contributing! I have discussed this with @skatsaounis and think the idea for this data source is justified, but I may suggest some changes. 

We don't need to read in all attributes, as we don't want users to rely heavily on this resource, as it could become quite heavy weight as the number of machines grows. To get around this, I suggest dropping most attributes and just keeping the essentials: `hostname` and `system_id` of the machine. Only if desired, using `for_each`, it's possible to get the full attributes with something like below: 

```HCL
data "maas_machines" "test" {}

data "maas_machine" "test" {
  for_each = data.maas_machines.test.machines
  hostname = each.key.hostname 
}
```
Or I believe as `for_each` only works with simple maps of strings, you may actually need to use `locals`: 
```HCL
# All machines as a type set of objects
data "maas_machines" "test" {}

# You may have to create a string map to use the for_each argument
locals {
  machines_map = {
    for machine in data.maas_machines.test.machines :
    machine.system_id => machine.hostname
  }
}

# Create machines if required
data "maas_machine" "test" {
  for_each = local.machines_map
  hostname = each.value 
}
```

Which you could use to get all machine attributes. 

This would be great as it future proofs us for upcoming changes to the gomaasclient. Let me know if you are happy with this, if so I can wait for your changes before doing a more thorough review.

maas/data_source_maas_machines.go
ricardbejarano: I've made the changes.

Truth is I believe Terraform providers should just mirror whatever the underlying API returns (and if that is the full set of attributes, then so be it), on the basis that this is not our choice to make but the API designers'; but my intended use case needs only `hostname`, so I'll just go with it.

ricardbejarano:
@tobiasdemendonca this is ready for review.

maas/data_source_maas_machines.go
```diff
@@ -0,0 +1,61 @@
...
+		items = append(items, map[string]interface{}{
+			"system_id": machine.SystemID,
+			"hostname":  machine.Hostname,
+		})
+	}
+
+	if err := d.Set("machines", items); err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(machines[0].SystemID)
```
skatsaounis, line 58: Let's just set something static here like `maas_machines` or `machines`

maas/resource_maas_machine_test.go
```diff
@@ -0,0 +1,85 @@
+package maas_test
```
skatsaounis, line 1: This is not needed at all. Please remove the entire file. Or more precisely, we have to add `maas_machine` acceptance tests but for the means of `maas_machines` data source they are not necessary. The reasoning is explained in the comment above.

maas/data_source_maas_machines_test.go
```diff
@@ -0,0 +1,49 @@
...
+	})
+}
+
+func testAccDataSourceMAASMachines(domain string, hostname string, power_parameters string, power_type string, pxe_mac_address string, zone string) string {
+	return fmt.Sprintf(`
+%s
+
+data "maas_machines" "test" {
+}
+`, testAccMAASMachine(domain, hostname, power_parameters, power_type, pxe_mac_address, zone))
```
skatsaounis, line 48: let's just keep only the datasource and return something like this:

```go
return `data "maas_machines" "test" {}`
```

maas/data_source_maas_machines_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testutils.PreCheck(t, nil) },
+		Providers:    testutils.TestAccProviders,
+		CheckDestroy: testAccCheckMAASMachineDestroy,
+		ErrorCheck:   func(err error) error { return err },
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceMAASMachines(domain, hostname, power_parameters, power_type, pxe_mac_address, zone),
+				Check:  resource.ComposeTestCheckFunc(checks...),
```
skatsaounis, line 36: Let's refactor the test to be something that fetches all the machines visible to the user with the gomaasclient, then confirm that all of them are present in the datasource machines field. This can be done with a pseudo code like this:

```go
conn := testutils.TestAccProvider.Meta().(*maas.ClientConfig).Client

# get all machines
machines, err := conn.Machines.Get(nil)

# confirm that machines equals what terraform knows
len(machines) == strconv.Atoi(rs.Primary.Attributes["machines.#"])
for _, i := range strconv.Atoi(rs.Primary.Attributes["machines.#"]) -1 {
    hostname := rs.Primary.Attributes[fmt.Sprintf("machines.%v.hostname", i)]
    systemID := rs.Primary.Attributes[fmt.Sprintf("machines.%v.system_id", i)]
    found := false
    for machine, _ := range machines {
        if machine.SystemID == systemID && machine.hostname == machine.Hostname {
            found = true
            break
        }
    }
    if not found {
        # error
    }
}
```

docs/data-sources/machines.md
```diff
@@ -0,0 +1,48 @@
...
+
+- `id` (String) The ID of this resource.
+- `machines` (Set of Object) A set of machines visible to the user. (see [below for nested schema](#nestedatt--machines))
+
+<a id="nestedatt--machines"></a>
+### Nested Schema for `machines`
+
+Read-Only:
+
+- `architecture` (String)
```
skatsaounis, line 41: this one needs `make generate_docs` rerun

examples/data-sources/maas_machines/data-source.tf
```diff
@@ -0,0 +1,10 @@
+resource "maas_machine" "virsh_vm1" {
```
skatsaounis, line 1: the resource is not needed here

skatsaounis (changes requested):
Hi @ricardbejarano 

Thank you for your PR. Please take a look at my comments. I will be more than happy to re-review, answer any question and eventually approve the changes. :slightly_smiling_face:

maas/data_source_maas_machines.go
ricardbejarano: Will do

ricardbejarano:
A couple of notes on the test refactor.

1. The method you're suggesting introduces a race condition. We're running `client.Machines.Get` twice. If the list of machines in the MAAS instance changes in between the two times we look, the test will fail for no fault of their own.
I don't know if your CI MAAS instance (if any) is read-only –or treated as such, at least on the Machines API–, but that's one option to prevent this.
I'll let you decide whether this is OK going forward or if further action is needed to prevent this.
My opinion is that this testing is significantly better than what I copy-pasted from other data sources in my first commits. This issue is worth noting (maybe in code comments? let me know), but a `make test` rerun should make it go away, as the chances of changes going in between the milliseconds between both reads, twice, are very slim.

1. I've skipped your suggestion of using search loops to compare both slices of machines, reason being the Machines API returns machines sorted by `entity.Machine.SystemID` (according to the MAAS API reference), and the `gomaasclient` doesn't mess with the original order.

1. `testutils.TestAccProvider.Meta()` is only non-nil from _within_ the test function (after `provider.Configure` has been called, so we must pack it all into a custom `resource.TestCheckFunc`.
And, since accessing state by address is only available in the `resource` packages' built-in functions, I've decided to manually invoke them by passing state and returning their error, if non-nil.
While perhaps less performant (haven't tested), it's simpler to read and I've preferred to optimize for that.


Let me know if additional changes are required. Also, let me know if you need me to make all these changes to [my other PR](https://github.com/canonical/terraform-provider-maas/pull/298) for `maas_devices`, so I can get those ready in parallel while you review.

skatsaounis (commented):
Hi @ricardbejarano 

The changes LGTM. Before I sent it to our CI let's try to add the final touches:

- Please rebase to the latest master
- Please add a comment on top of the for loop to refer to the behavior of the API to always return the machines sorted by their `system_id`.
- Please add a comment on the testcase, most probably included in the length test failure message, indicating that there is chance the overall machines have changed, so the future developer should try to re-run the test case.

Let me know when this is done so that I will trigger a CI run. I suggest we wait for the merge of this PR before updating the one related to devices.

Thanks a lot for addressing all my comments, and of course for the contribution!

ricardbejarano:
@skatsaounis done!

maas/data_source_maas_machines_test.go
```diff
@@ -0,0 +1,60 @@
...
+	checks := []resource.TestCheckFunc{
+		func(s *terraform.State) error {
+			conn := testutils.TestAccProvider.Meta().(*maas.ClientConfig).Client
+
+			machines, err := conn.Machines.Get(nil)
+			if err != nil {
+				return err
+			}
+
+			// We're about to compare the `machines` we just got, with the `machines` data source attribute.
```
skatsaounis, line 24: Our CI failed because we run all acceptance tests together and we create a new VM (machine) as part of `maas_vm_host` acceptance tests. I should have guessed that. Let's make the test simpler, as in the end, Terraform provider resource does not care about the order of the machines. In addition, what we need to check is that for each machine the `system_id` and `hostname` are set. A diff like the below could do the trick. Worst case, we could still fetch the machines with the gomaasclient and try to confirm that for every `system_id` and `hostname` that Terraform knows, this combination can also be found in the gomaasclient response. I am happy to approve though without this check. Please re-update the PR and hopefully our CI will give a :+1: 

```diff
-			machines, err := conn.Machines.Get(nil)
-			if err != nil {
-				return err
-			}
-
-			// There's a slim chance that the upstream MAAS API list of machines changed in between both times we read.
-			// If any of the following tests fail for no apparent reason, that might explain why: a change might've gone through in between them.
-			// Try re-running the tests if you suspect that's what's happening.
-			if err := resource.TestCheckResourceAttr("data.maas_machines.test", "machines.#", strconv.Itoa(len(machines)))(s); err != nil {
-				return err
-			}
-			// According to MAAS' API documentation, machines are returned "sorted by id (i.e. most recent last)"
-			for i, machine := range machines {
+			for i, machine := range strconv.Atoi(rs.Primary.Attributes["machines.#"]) {
-				if err := resource.TestCheckResourceAttr("data.maas_machines.test", fmt.Sprintf("machines.%v.system_id", i), machine.SystemID)(s); err != nil {
+				if err := resource.TestCheckResourceAttrSet("data.maas_machines.test", fmt.Sprintf("machines.%v.system_id", i))(s); err != nil {
					return err
				}
-				if err := resource.TestCheckResourceAttr("data.maas_machines.test", fmt.Sprintf("machines.%v.hostname", i), machine.Hostname)(s); err != nil {
+				if err := resource.TestCheckResourceAttrSet("data.maas_machines.test", fmt.Sprintf("machines.%v.hostname", i))(s); err != nil {
					return err
				}
			}
```

skatsaounis (changes requested):

ricardbejarano:
@skatsaounis done!

Issue #189, opened 2024-06-11
Feature Request: Add storage device details of the enlisted machine to Terraform state during commissioning


vishvikkrishnan:
Currently, the `maas_machine` resource does not provide details about the storage devices present in the commissioned node. This is unlike network interfaces, which are auto-populated with the MAC addresses of each NIC discovered in the enlisted machine.
```hcl
# maas_machine.node["node01ob48"] will be created
  + resource "maas_machine" "node" {
      + architecture       = "amd64/generic"
      + domain             = (known after apply)
      + hostname           = "node01ob48"
      + id                 = (known after apply)
      + min_hwe_kernel     = (known after apply)
      + network_interfaces = (known after apply)
      + pool               = (known after apply)
      + power_parameters   = (sensitive value)
      + power_type         = "amt"
      + pxe_mac_address    = "b8:ae:ed:73:38:40"
      + zone               = "default"
    }
```
After applying the above resource, the following details are available in the TF state:

```hcl
# maas_machine.node["node01ob48"]:
resource "maas_machine" "node" {
    architecture       = "amd64/generic"
    domain             = "maas"
    hostname           = "node01ob48"
    id                 = "c7xnp8"
    min_hwe_kernel     = null
    network_interfaces = [
        "8c:ae:4c:df:50:7c",
        "b8:ae:ed:73:38:40",
    ]
    pool               = "default"
    power_parameters   = (sensitive value)
    power_type         = "amt"
    pxe_mac_address    = "b8:ae:ed:73:38:40"
    zone               = "default"
}
```

However, the state file does not track the storage device details, unlike network interface details that are auto-populated. MAAS automatically enlists all storage devices during node enlistment and, by default, selects and partitions the first device from the alphanumerically sorted storage disk list as its boot and / partitions: 
![image](https://github.com/maas/terraform-provider-maas/assets/44338064/c7f93006-3ccd-4546-86b3-69e0753f41b9)

Because of this, trying to apply the `maas_block_device` resource with a different target disk causes an error:
```hcl
# maas_block_device.disks["node01ob48-sdb"] will be created
  + resource "maas_block_device" "disks" {
      + block_size     = 512
      + id             = (known after apply)
      + id_path        = (known after apply)
      + is_boot_device = false
      + machine        = "node01ob48"
      + model          = "KINGSTON SM2280S"
      + name           = "sdb"
      + path           = (known after apply)
      + serial         = "50026B725A01A663"
      + size_gigabytes = 120
[...]

PR #307, opened 2025-04-16
feat: 189-storage-device-details-of-enlisted-machines

tobiasdemendonca:
## Description of changes

Add storagee device details to `maas_machine`. 

Resolves: #189

## Checklist


maas/resource_maas_machine.go
```diff
@@ -68,6 +69,35 @@ func resourceMAASMachine() *schema.Resource {
...
+			"block_devices": {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Description: "A list of block devices attached to the machine.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id_path": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "The ID path of the block device. This should be a path that is fixed and doesn't change depending on the boot order or kernel version.",
```
skatsaounis, line 81: The description here can confuse the user since this field is computed. "This should be a path ....". Most probably it was copied from `maas_block_device` resource.

maas/resource_maas_machine.go
```diff
@@ -68,6 +69,35 @@ func resourceMAASMachine() *schema.Resource {
...
+						},
+						"model": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "Model of the block device. Used in conjunction with `serial` argument. Conflicts with `id_path`. This argument is computed if it's not given.",
```
skatsaounis, line 86: Same thing here. Let's try to change both descriptions so that they describe the field but as something computed, not specified by the user.

skatsaounis (changes requested):
I added some comments for improving the description of the fields. In addition, I do not know if it is a bug or not, but I spotted that the Markdown Does not contain descriptions of the read-only fields.

Finally, given that we do not have yet a `maas_machine` acceptance test we have to decide whether it is a good time to introduce one with this PR or avoid it.

tobiasdemendonca:
> I added some comments for improving the description of the fields. In addition, I do not know if it is a bug or not, but I spotted that the Markdown Does not contain descriptions of the read-only fields.
> 
> Finally, given that we do not have yet a `maas_machine` acceptance test we have to decide whether it is a good time to introduce one with this PR or avoid it.

I thought we discussed not including the maas machine test for this PR but I could be wrong. I do think we probably need one, but I guess the setup would be a bit complex if I were to mimic what we discussed a few days ago which involved creating a vm, renaming it outside of maas, removing the old resource from maas, and then using the `maas_machine` to create and commission it again. 

I might suggest adding that in a separate PR because of this, what do you think? Ultimately though I can prioritise this sprint or next sprint if we think it's really required: just because it's a bit harder to implement doesn't mean we shouldn't add it! But I think it could be considered related but separate to this PR.

tobiasdemendonca:
Okay I found an existing issue, doesn't look like this will be fixed soon https://github.com/hashicorp/terraform-plugin-docs/issues/28 I can try and work out how to add this in manually

skatsaounis (approved):
No reason to add things manually for now. The PR LGTM +1. Try to rebase, then make the CI happy and finally merge it :beers:

tobiasdemendonca:
I believe the best way is with a template, I implemented it but as soon as you want to adjust the schema, you have to write it manually so there's a risk of it getting out of sync. I've reverted my changes for now but can discuss this when we're back if needed

SK1Y101 (dismissed):
+1 from me too

skatsaounis (approved):
LGTM +1

Issue #260, opened 2025-02-11
terraform fail to destroy maas_instance when disk erase set to default in MAAS


noama-nv:
PSB expected status after release is the machine to be ready seems like any other status considered as bad
please add this stage as optional in transitioning from deployed to Ready so that the delete timeout is really effective

```
maas_instance.controllers[0]: Destroying... [id=b4qqxb]
maas_instance.workers[0]: Destroying... [id=6bwhkd]
maas_instance.workers[1]: Destroying... [id=4pdmwt]
maas_instance.workers[1]: Still destroying... [id=4pdmwt, 10s elapsed]
maas_instance.workers[0]: Still destroying... [id=6bwhkd, 10s elapsed]
maas_instance.controllers[0]: Still destroying... [id=b4qqxb, 10s elapsed]
maas_instance.controllers[0]: Destruction complete after 11s
maas_instance.workers[0]: Still destroying... [id=6bwhkd, 20s elapsed]
maas_instance.workers[1]: Still destroying... [id=4pdmwt, 20s elapsed]
maas_instance.workers[0]: Still destroying... [id=6bwhkd, 30s elapsed]
maas_instance.workers[1]: Still destroying... [id=4pdmwt, 30s elapsed]
maas_instance.workers[1]: Still destroying... [id=4pdmwt, 40s elapsed]
maas_instance.workers[0]: Still destroying... [id=6bwhkd, 40s elapsed]
maas_instance.workers[1]: Still destroying... [id=4pdmwt, 50s elapsed]
maas_instance.workers[0]: Still destroying... [id=6bwhkd, 50s elapsed]
╷
│ Error: unexpected state 'Disk erasing', wanted target 'Ready'. last error: %!s(<nil>)
│ 
│ 
╵
╷
│ Error: unexpected state 'Disk erasing', wanted target 'Ready'. last error: %!s(<nil>)
│ 
│ 
╵
Releasing state lock. This may take a few moments...
```

Issue #72, opened 2023-06-10
Feature Request: Option to Erase Disks when destroying instances.


vyas-n:
When releasing machines in the MaaS UI, users can choose to erase disks.

I'd like for a way specify this when destroying maas_instance terraform resources.

PR #310, opened 2025-04-23
feat: 72-erase-disks-when-destroying-instances

tobiasdemendonca:
## Description of changes

Add ability to erase disks when destroying the resource `maas_instance`. 

Resolves: #72 
Resolves: #260

## Checklist

### DIFF
--- a/docs/resources/instance.md
+++ b/docs/resources/instance.md
@@ -42,6 +42,7 @@ resource "maas_instance" "kvm" {
 - `allocate_params` (Block List, Max: 1) Nested argument with the constraints used to machine allocation. Defined below. (see [below for nested schema](#nestedblock--allocate_params))
 - `deploy_params` (Block List, Max: 1) Nested argument with the config used to deploy the allocated machine. Defined below. (see [below for nested schema](#nestedblock--deploy_params))
 - `network_interfaces` (Block Set) Specifies a network interface configuration done before the machine is deployed. Parameters defined below. This argument is processed in [attribute-as-blocks mode](https://www.terraform.io/docs/configuration/attr-as-blocks.html). (see [below for nested schema](#nestedblock--network_interfaces))
+- `release_params` (Block List, Max: 1) Parameters used to release the allocated machine. Only used when the machine is released upon resource destroy. (see [below for nested schema](#nestedblock--release_params))
 - `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
 
 ### Read-Only
@@ -95,6 +96,18 @@ Optional:
 
 **NOTE:** If both `subnet_cidr` and `ip_address` are not defined, the interface will not be configured on the allocated machine.
 - `subnet_cidr` (String) An existing subnet CIDR used to configure the network interface. Unless `ip_address` is defined, a free IP address is allocated from the subnet.
+
+
+<a id="nestedblock--release_params"></a>
+### Nested Schema for `release_params`
+
+Optional:
+
+- `comment` (String) A comment to be added to the event log when the machine is released.
+- `erase` (Boolean) Erase the disk when releasing.
+- `force` (Boolean) Force the release of the machine. If the machine was deployed as a KVM host, all machines inside the host will be deleted. Use with caution.
+- `quick_erase` (Boolean) Use quick erase. Wipe 2MiB at the start and at the end of the drive to make data recovery inconvenient and unlikely to happen by accident. This is not secure.
+- `secure_erase` (Boolean) Use the drive's secure erase feature if available.  In some cases, this can be much faster than overwriting the drive. Some drives implement secure erasure by overwriting themselves so this could still be slow.
 
 
 <a id="nestedblock--timeouts"></a>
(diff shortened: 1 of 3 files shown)


tobiasdemendonca:
Final check of the test is currently failing

maas/resource_maas_instance_test.go
```diff
@@ -0,0 +1,192 @@
...
+					resource.TestCheckResourceAttr("maas_instance.test", "release_params.0.quick_erase", quickErase),
+					resource.TestCheckResourceAttr("maas_instance.test", "release_params.0.secure_erase", secureErase),
+				)...,
+				),
+			},
+			// Test destroy leaves the machine in a ready state
+			{
+				Config: testAccMAASInstanceConfigSetup(vmHost, hostname),
+				Check:  testAccMAASInstanceCheckMachineInStatus(machine.SystemID, node.StatusReady),
+			},
```
tobiasdemendonca, line 68: Can't think of a better way of checking if release has been called with the correct parameters. `client.Events.Get` does not show the log comments I was hoping for

tobiasdemendonca:
> Final check of the test is currently failing

Fixed

docs/resources/instance.md
```diff
@@ -97,6 +98,18 @@ Optional:
...
 
 
+<a id="nestedblock--release_params"></a>
+### Nested Schema for `release_params`
+
+Optional:
+
+- `comment` (String) A comment to be added to the event log when the machine is released.
+- `erase` (Boolean) Erase the disk when releasing.
+- `force` (Boolean) Force the release of the machine. If the machine was deployed as a KVM host, all machines inside the host will be deleted. Use with caution.
```
tobiasdemendonca, line 108: I guess the machine will never be deployed as the kvm host with this resource, so I will remove this part of the description with the rest of the review comments

tobiasdemendonca: Note that descriptions were taken from https://maas.io/docs/machine

maas/resource_maas_instance.go
```diff
@@ -283,17 +319,24 @@ func resourceInstanceRead(ctx context.Context, d *schema.ResourceData, meta any)
 	return nil
 }
 
+func resourceInstanceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
```
tobiasdemendonca, line 322: Have followed the pattern already existing in this file, but this deviates from the norm. Was this for any particular reason why this isn't `resourceMAASInstanceUpdate`?

maas/resource_maas_instance.go
```diff
@@ -197,6 +198,41 @@ func resourceMAASInstance() *schema.Resource {
...
+							Type:        schema.TypeString,
+							Optional:    true,
+							Description: "A comment to be added to the event log when the machine is released.",
+						},
+						"erase": {
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Description: "Erase the disk when releasing.",
+						},
+						"force": {
```
SK1Y101, line 218: I know the MAAS API has this as force, but i'm wondering whether `force_release` or something slightly more informative might be useful for the terraform side @skatsaounis ?

maas/resource_maas_instance.go
```diff
@@ -197,6 +198,41 @@ func resourceMAASInstance() *schema.Resource {
...
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Description: "Force the release of the machine. If the machine was deployed as a KVM host, all machines inside the host will be deleted. Use with caution.",
```
SK1Y101, line 221: also yeah of course slightly update if this won't be a KVM host, but you've covered this already above

maas/resource_maas_instance_test.go
```diff
@@ -0,0 +1,211 @@
...
+)
+
+func TestAccResourceMAASInstance_basic(t *testing.T) {
+	vmHost := os.Getenv("TF_ACC_VM_HOST_ID")
+	hostname := acctest.RandomWithPrefix("tf-instance")
+	comment := acctest.RandomWithPrefix("tf-instance-comment")
+	erase := "true"
+	force := "false"
+	quickErase := "true"
+	secureErase := "false"
```
SK1Y101, line 26: won't this violate the `MaxItems: 1` constraint on the release params, or am I misunderstanding something?

actually i think I've realised, that allows you to only specify release params once per machine, rather than only include one of the sub-fields at a time?

maas/resource_maas_instance_test.go
```diff
@@ -0,0 +1,211 @@
...
+		wasReleased := false
+
+		for _, event := range events.Events {
+			if event.Type == "Disks erased" {
+				wasErased = true
+			}
+
+			if event.Type == "Released" {
+				wasReleased = true
+			}
```
SK1Y101, line 98: you could probably also exit early here with something like `if (wasReleased && wasErased) { break }`
Not that there will be a huge performance hit probably if you do look at the rest of the event log

SK1Y101 (dismissed):
Awesome! a few questions clarifying my understanding, but looks good!!

maas/resource_maas_instance_test.go
tobiasdemendonca: Yes once per `maas_instance` resource, see `deploy_params` or `allocate_params` in this resource as an example

maas/resource_maas_instance_test.go
tobiasdemendonca: I initially agreed but actually realised I have to check the entire log, because if the machine has been erased when it shouldn't be expected, then this should throw an error. 

Thinking about this more thoroughly did catch an additional error in this test which I have just fixed

maas/resource_maas_instance.go
skatsaounis: I have no strong preference here. Typically, I follow the API so that users that using both are not confused.

skatsaounis (approved):
LGTM +1

PR #323, opened 2025-05-27
refactor: Update user resource

SK1Y101:
## Description of changes

Adds the missing user fields to the user endpoints, and implements testing for the user resource.

Users may now define a `transfer_to_user` resource, as per MAAS user delete.
Additionally the field `is_local` is populated, as per MAAS return.

User read now correctly sets the terraform state of the object, as per current practices.

Requires: https://github.com/canonical/gomaasclient/pull/134

(implements the missing user fields at the gomaasclient layer)

## Checklist

- [x] I have written a PR title that follows the advice in DEVELOPMENT.md with the title format `type: title`

### DIFF
--- a/maas/resource_maas_user.go
+++ b/maas/resource_maas_user.go
@@ -29,6 +29,7 @@ func resourceMAASUser() *schema.Resource {
 					"name":     user.UserName,
 					"email":    user.Email,
 					"is_admin": user.IsSuperUser,
+					"is_local": user.IsLocal,
 				}
 				if err := setTerraformState(d, tfState); err != nil {
 					return nil, err
@@ -52,6 +53,12 @@ func resourceMAASUser() *schema.Resource {
 				ForceNew:    true,
 				Description: "Boolean value indicating if the user is a MAAS administrator. Defaults to `false`.",
 			},
+			"is_local": {
+				Type:        schema.TypeBool,
+				Default:     true,
+				Computed:    true,
+				Description: "Boolean value indicating if the user is a local MAAS account. Defaults to `true`.",
+			},
 			"name": {
 				Type:        schema.TypeString,
 				Required:    true,
@@ -64,6 +71,11 @@ func resourceMAASUser() *schema.Resource {
 				Sensitive:   true,
 				ForceNew:    true,
 				Description: "The user password.",
+			},
+			"transfer_to_user": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "If provided, resources owned by the deleted user will be transfered to this user. ",
 			},
 		},
 	}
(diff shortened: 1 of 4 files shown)


SK1Y101:
investigate a custom validator that would ensure you cannot manage yourself at terraform (using whoami)

skatsaounis (changes requested):
Thank you for the work. Some things identified during review:

- It is nice that we have a check so that the API key owner cannot be managed as a user. It is better though to have the check at the import and create functions only. If we do so, the check is not needed in delete
- is_local adds no value to resource `maas_user` since Terraform can only create local users. What makes sense though, is to prevent importing non-local users.

maas/resource_maas_user.go
```diff
@@ -12,7 +12,7 @@ import (
 
 func resourceMAASUser() *schema.Resource {
 	return &schema.Resource{
-		Description:   "Provides a resource to manage MAAS users.",
+		Description:   "Provides a resource to manage MAAS users. *Note* You cannot use this to modify the logged in terraform user.",
```
skatsaounis, line 15: Maybe it makes sense to refer to non-local users that also cannot be managed by Terraform

maas/resource_maas_user.go
```diff
@@ -65,37 +69,75 @@ func resourceMAASUser() *schema.Resource {
 				ForceNew:    true,
 				Description: "The user password.",
 			},
+			"transfer_to_user": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "If provided, resources owned by the deleted user will be transfered to this user. ",
```
skatsaounis, line 76:
```suggestion
				Description: "If provided, resources owned by the deleted user will be transferred to this user.",
```

maas/resource_maas_user.go
```diff
@@ -125,3 +167,22 @@ func getUser(client *client.Client, userName string) (*entity.User, error) {
 
 	return nil, fmt.Errorf("user (%s) was not found", userName)
 }
+
+func verifyUserValid(client *client.Client, d *schema.ResourceData) error {
+	// ensure the user is a valid target for create/delete
```
skatsaounis, line 172:
```suggestion
	// ensure the user is a valid target for create/import
```

maas/resource_maas_user.go
```diff
@@ -125,3 +167,22 @@ func getUser(client *client.Client, userName string) (*entity.User, error) {
...
+	if err != nil {
+		return err
+	}
+
+	name := d.Get("name").(string)
+	if name == me.UserName {
+		return fmt.Errorf("cannot operate on the currently logged in user %q", me.UserName)
+	}
+
+	if !me.IsLocal {
```
skatsaounis, line 183: The check here is wrong. Thinking about it more, the check should only happen at import. Reasoning: if Terraform tries to create a non local user or the same user as the owner of the api_key MAAS will throw an error either way. In both cases, a user with that username or email already exists. So there is no need to check during create.

For import the check is a different one than what this function is doing:

We have to confirm that there is a user in MAAS with the `d.Id()`, and then check that the user is not me (whoami) or a non local user. Pseudocode

```go
user, err := getUser(client, d.Id())
me, err := client.Users.Whoami()

if user.IsLocal {error}
if user.UserName == me.UserName {error}
```

skatsaounis (changes requested):
I spotted some extra issues, plus there is a need for recreating the markdown, as detected by the CI

Issue #291, opened 2025-03-26
feat: management of MAAS configuration with Terraform


skatsaounis:
Add a Terraform resource to manage MAAS global configuration values and a supplementary data source for retrieving applied configuration.

Resource examples:

```tf
# Alt 1 - set the value of a specific configuration key
resource "maas_configuration" "setting_one" {
  key    = "tls_cert_expiration_notification_interval"
  value = "30" 
}

resource "maas_configuration" "setting_two" {
  key    = "maas_auto_ipmi_cipher_suite_id"
  value = "3" 
}
```
```tf
# Alt 2 - set the value of many configuration keys (strict or loose check on the key name)
resource "maas_configuration" "settings" {
  tls_cert_expiration_notification_interval = "30"
  maas_auto_ipmi_cipher_suite_id            = "3"
}
```

Data source examples:

```tf
# Alt 1 - request the value of a specific configuration key
data "maas_configuration" "setting_one" {
  key    = "tls_cert_expiration_notification_interval"
}

data "maas_configuration" "setting_two" {
  key    = "maas_auto_ipmi_cipher_suite_id"
}
```
```tf
# Alt 2 - get all known configuration keys and their values
data "maas_configuration" "settings" {}
```

PR #326, opened 2025-06-02
feat: add maas_configuration resource

tobiasdemendonca:
## Description of changes

Add the `maas_configuration` resource to control maas settings as key value pairs.

```hcl
resource "maas_configuration" "test_bool" {
  key   = "release_notifications"
  value = "true"
}
```

Resolves: #291

## Checklist

- [x] I have written a PR title that follows the advice in DEVELOPMENT.md with the title format `type: title`


tobiasdemendonca:
@skatsaounis here are the tests so far. Happy to remove some test cases to just check every type, but note that I wouldn't have caught the fact that `remote_syslog` test fails because when set to an empty string, MAAS returns `null` instead of `""`

examples/resources/maas_configuration/import.sh
```diff
@@ -0,0 +1,2 @@
+# Import a MAAS configuration setting using the key as the identifier
+terraform import maas_configuration.test release_notifications
```
skatsaounis, line 2: the usual suspect nit: Please add a new line here and re-create docs. If we end up removing import functionality, feel free to delete the file instead.

maas/resource_maas_configuration.go
```diff
@@ -0,0 +1,103 @@
...
+		UpdateContext: resourceMAASConfigurationUpdate,
+		DeleteContext: resourceMAASConfigurationDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: map[string]*schema.Schema{
+			"key": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
```
skatsaounis, line 29: Why do we need a `ForceNew` here?

maas/resource_maas_configuration.go
```diff
@@ -0,0 +1,103 @@
...
+
+func resourceMAASConfiguration() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Provides a resource to manage MAAS configuration settings through key value pairs. See MAAS server in the [MAAS API documentation](https://maas.io/docs/api) for a full list of keys and their values. Upon destroy, the specified key will no longer be managed by Terraform, but the set value will remain set in MAAS.",
+		CreateContext: resourceMAASConfigurationCreate,
+		ReadContext:   resourceMAASConfigurationRead,
+		UpdateContext: resourceMAASConfigurationUpdate,
+		DeleteContext: resourceMAASConfigurationDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
```
skatsaounis, line 23: I doubt that we need an import functionality. This is a static value that we set, and we can see with a Terraform state refresh which is the MAAS value for it vs the user's Terraform plan. Most probably we can skip the import.

maas/resource_maas_configuration.go
```diff
@@ -0,0 +1,103 @@
...
+		},
+		Schema: map[string]*schema.Schema{
+			"key": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				Description:      "Key corresponding to the configuration setting.",
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringDoesNotMatch(regexp.MustCompile("maas_auto_ipmi_workaround_flags"), "Key 'maas_auto_ipmi_workaround_flags' cannot currently be set through Terraform due to a [bug in MAAS](https://bugs.launchpad.net/maas/+bug/2112191). A fix will be worked on for a future MAAS releases.")),
```
skatsaounis, line 31: Since there is chance that we will not fix it at all, there is no need to hard-code this message in the provider's messages:

```suggestion
				ValidateDiagFunc: validation.ToDiagFunc(validation.StringDoesNotMatch(regexp.MustCompile("maas_auto_ipmi_workaround_flags"), "Key 'maas_auto_ipmi_workaround_flags' cannot currently be set through Terraform due to a [bug in MAAS](https://bugs.launchpad.net/maas/+bug/2112191).")),
```

maas/resource_maas_configuration_test.go
```diff
@@ -0,0 +1,462 @@
...
+			if res != testCase.expected {
+				t.Errorf("expected %q, got %q for case %q", testCase.expected, res, testCase.label)
+			}
+		})
+	}
+}
+
+func TestAccResourceMAASConfiguration_basic(t *testing.T) {
+	// Test all MAAS Settings cases. Set twice to ensure a change is actually made
+	// incase defaults change, and value 2 is selected to be a typical value to
```
skatsaounis, line 59:
```suggestion
	// in case defaults change, and value 2 is selected to be a typical value to
```

maas/resource_maas_configuration_test.go
```diff
@@ -0,0 +1,462 @@
...
+		},
+		{
+			key:    "boot_images_no_proxy",
+			value1: "true",
+			value2: "false",
+		},
+		{
+			key:    "commissioning_distro_series",
+			value1: "noble",
+			value2: "noble", // Dev environment does not necessarily have multiple distros
```
skatsaounis, line 89: there is no guarantee that noble is available at the MAAS under test. What about using a similar approach to `TF_ACC_BOOT_RESOURCES_OS` for boot resources? The CI should be updated accordingly to provide to acceptance tests the needed environment variable.

maas/resource_maas_configuration_test.go
```diff
@@ -0,0 +1,462 @@
...
+		},
+		{
+			key:    "default_boot_interface_link_type",
+			value1: "static",
+			value2: "auto",
+		},
+		{
+			key:    "default_distro_series",
+			value1: "noble",
+			value2: "noble", // Dev environment does not necessarily have multiple distros
```
skatsaounis, line 109: Same as the above comment for OS availability at MAAS under test.

skatsaounis (changes requested):
Thanks for the PR. Please take a look at the comments. We are almost good to go :slightly_smiling_face:

skatsaounis:
> This adds about 45s to the acceptance tests, with all the new maas_configuration tests running in parallel on my local machine.
> 
> It's not a significant overhead, but we could reduce this to test every value _type_ instead. I wouldn't have caught the issue with `remote_syslog` without having tested everything, so it could be worth keeping all of them, but I'm interested to hear what anyone thinks

I am super ok to test all the known configuration options as of today. No need to reduce the tests.

maas/resource_maas_configuration.go
tobiasdemendonca: The key is used as the resource id, so that you cannot have multiple resources managing the same settings. If this key is changed, the id will change, meaning it's effectively a new resource, so on principle this should be force new. 

Because destroy does nothing, this doesn't actually matter, but I _think_ this is correct

skatsaounis (approved):
LGTM +1

tobiasdemendonca:
@skatsaounis this has (finally) passed ci, but is dependent on this MP landing first: https://code.launchpad.net/~tobydm/maas-ci/+git/system-tests/+merge/486549 . Would you mind reviewing it when you get the chance please?

SK1Y101 (approved):
awesome, looks good to me!

SK1Y101 (approved):
Awesome!!! +1

Issue #343, opened 2025-07-15
performance problem when using resources of type maas_block_device


pescobar:
We have noticed a huge increase in the execution time of `tofu plan` and  `tofu apply` since we started adding resources of type `maas_block_device`

As example, this is the wallclock time for a `tofu plan` of a module managing 40 physical machines:

```
time tofu plan -target='module.scb_nodes_compute'

No changes. Your infrastructure matches the configuration.

real    0m15.287s
user    0m7.992s
sys     0m1.438s
```

And this is the execution for another module with only 4 physical machines. The definition is the same but we have added a resource of type `maas_block_device` to each of them

```
time tofu plan -target='module.pcd_nodes_tofu_compute'

No changes. Your infrastructure matches the configuration.

real    0m30.364s
user    0m8.845s
sys     0m1.618s
```

A module with 10 times less machines takes double the time to execute `tofu plan` . Our issue is that we would like to define `maas_block_device` for hundred of machines so it won't scale and each execution of opentofu would take too long.

When executing the `tofu plan` you can easily identify it gets stuck most of the time refreshing the state of the block devices 

```
module.pcd_nodes_tofu_compute.maas_block_device.pcd_sda["pcd01"]: Refreshing state... [id=2644]
module.pcd_nodes_tofu_compute.maas_block_device.pcd_sda["pcd04"]: Refreshing state... [id=2642]
module.pcd_nodes_tofu_compute.maas_block_device.pcd_sda["pcd03"]: Refreshing state... [id=2643]
module.pcd_nodes_tofu_compute.maas_block_device.pcd_sda["pcd02"]: Refreshing state... [id=2641]
```

When reviewing the opentofu trace logs you can also notice most of the time is spent in the block_device related tasks (notice they start at `13:02:25` and finish at `13:02:40`

```
grep sda ~/tmp/tofu-trace-pcd-nodes.log |egrep 'start|complete'
2025-07-15T13:02:25.366+0200 [TRACE] vertex "module.pcd_nodes_tofu_compute.maas_block_device.pcd_sda (expand)": starting visit (*tofu.nodeExpandPlannableResource)
2025-07-15T13:02:25.367+0200 [TRACE] vertex "module.pcd_nodes_tofu_compute.maas_block_device.pcd_sda[\"pcd03\"]": starting visit (*tofu.NodePlannableResourceInstance)
[...]

PR #354, opened 2025-09-14
fix: getMachine defaults to system-id

jameinel:
## Description of changes

When trying to resolve an identifier into a Machine, first do a simple one-machine query to lookup the identifier by SystemID. Only once that fails do we fall back to the previous algorithm, which would read all machines, and do a one-by-one comparison to see if that was a SystemID, a Hostname, a FQDN, or the boot MACAddress.

A very common pattern with Terraform is going to be defining a Machine resource, and then referencing it later using `mymachine.id` and so we can do a single lookup on that attribute, rather than reading all machines for any reference.

Resolves: #343 

## Checklist

- [ ] I have written a PR title that follows the advice in DEVELOPMENT.md with the title format `type: title`


jameinel:
I have not yet written any tests, because I wanted to validate that it is worth pursuing this path and that it has a chance of being merged. But if you are happy with the direction, I can follow up with tests, etc.

SK1Y101:
I'm pretty happy with the idea of short circuiting the `getMachine` function with the machine SystemID, it's a common operation and the potential time and resource saves are significant.

If you can get some tests (and the linting) passing, I'd be happy to see this merged

tobiasdemendonca:
Hey @jameinel , to reiterate @SK1Y101, the team are keen to go with the addition of this extra call to improve performance. Can you let us know if you plan to continue to work on this, otherwise we will aim to take this PR over in the new year to get this across the line. Thanks again for the contribution!

goldberl:
Hi @tobiasdemendonca and @SK1Y101, just wanted to let you know that I am currently working with @jameinel to add relevant tests for this, thanks!

skatsaounis (changes requested):
Please fetch the latest master branch and also take a look at my inline comment. Otherwise, I am happy to approve the change.

skatsaounis (changes requested):

PR #384, opened 2025-11-14
feat: Add `architecture` field to MAAS instance resource

andylshort:
## Description of changes

Added the `architecture` field to the MAAS instance resource to support machines using different architectures.

Resolves: https://github.com/canonical/terraform-provider-maas/issues/236

## Checklist

### DIFF
--- a/maas/resource_maas_vlan_dhcp.go
+++ b/maas/resource_maas_vlan_dhcp.go
@@ -10,6 +10,7 @@ import (
 	"github.com/canonical/gomaasclient/entity"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
 func resourceMAASVLANDHCP() *schema.Resource {
@@ -47,6 +48,7 @@ func resourceMAASVLANDHCP() *schema.Resource {
 				Optional:      true,
 				ConflictsWith: []string{"primary_rack_controller", "secondary_rack_controller"},
 				AtLeastOneOf:  []string{"primary_rack_controller", "relay_vlan"},
+				ValidateFunc:  validation.IntAtLeast(1),
 				Description:   "Database ID of the VLAN to use as a relay for DHCP.",
 			},
 			"secondary_rack_controller": {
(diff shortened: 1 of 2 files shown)


maas/resource_maas_instance_test.go
```diff
@@ -71,6 +71,7 @@ func TestAccResourceMAASInstance_basic(t *testing.T) {
 		resource.TestCheckResourceAttr("maas_instance.test", "hostname", hostname),
 		resource.TestCheckResourceAttr("maas_instance.test", "memory", "4096"),
 		resource.TestCheckResourceAttr("maas_instance.test", "cpu_count", "1"),
+		resource.TestCheckResourceAttr("maas_instance.test", "architecture", "amd64/generic"),
```
tobiasdemendonca: This is only testing the `resourceInstanceRead` function which is reading the default value. You need to either have done some QA or write an additional test that allows you to test if specifying the arch is actually sent in the allocate params. 

One way could be to try to specify an architecture that does not exist, assuming maas will throw an error that you can check for.  Another could be actually allocate a machine of a different architecture, but that would be more involved from a ci perspective.

maas/resource_maas_instance.go
```diff
@@ -45,6 +45,13 @@ func resourceMAASInstance() *schema.Resource {
 				Description: "Nested argument with the constraints used to machine allocation. Defined below.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"architecture": {
```
SK1Y101: Do we want to restrict the architectures to the set MAAS allows, or rely on the api response to error if you supply, say `arm82`

maas/resource_maas_instance_test.go
SK1Y101: We don't have the capability to allocate other architectures in CI at the moment I believe, but it was something we wanted to look into.

Unfortunately that eliminates any obvious ways of `deploy machine, read machine details`, as any success will always be amd64 by necessity.

maas/resource_maas_instance.go
andylshort: Is there any way of getting this information ahead-of-time? This sounds like it needs to be an API call, one I'm not sure is publicly exposed or implemented. Maybe best to let the `terraform apply` fail due to the MAAS machine allocate failing as that's a guarantee, and leave a FIXME above?

tobiasdemendonca (approved):
LGTM with an optional nit

PR #407, opened 2026-01-21
fix: Corrected data validation in package repository resource

andylshort:
## Description of changes

This PR covers some data validation changes with regards to package repositories:

* Changed how empty lists are converted to strings from '[]' to '' to align with how the validation in MAAS expects these strings to be formed. Added an acceptance test to the resource, and some for the utility function that does this conversion, to reinforce behaviour.
* Converted `Enabled` from a boolean to a boolean pointer. This is due to `omitempty` in the JSON marshalling process ignoring `false` values, even if they're specified by the user. This prevented certain user-specified data from being passed to the MAAS API.
* Converted "package repositories" to "package repository" to make these consistent with the naming convention of the rest of the resources in the provider.

Resolves: https://github.com/canonical/terraform-provider-maas/issues/346

## Checklist

### DIFF
--- a/maas/resource_maas_package_repository_test.go
+++ b/maas/resource_maas_package_repository_test.go
@@ -192,7 +192,7 @@ resource "maas_package_repository" %q {
 }
 
 func testAccUbuntuPackageRepository(resourceName string, name string, key string, url string, disableSources bool, enabled bool, arches []string, disabledComponents []string, disabledPockets []string, distributions []string) string {
-	return fmt.Sprintf(`
+	resource := fmt.Sprintf(`
 resource "maas_package_repository" %q {
   name = %q
   key  = %q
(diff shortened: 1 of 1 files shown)


maas/utils_test.go
```diff
@@ -202,3 +202,36 @@ func TestCheckSemverConstraint(t *testing.T) {
...
+func TestListAsString(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    []any
+		expected string
+	}{
+		{
+			name:     "empty list",
+			input:    []any{},
+			expected: "",
```
tobiasdemendonca: Great to have these tests, if my earlier comment is correct then I think this is the test for listAsStringBase

skatsaounis (approved):

PR #463, opened 2026-05-19
refactor: Don't leave stale disks when renaming

### LINKED ISSUE #396: Setting Custom Disk Name with block_device resource would create stale disk definition
We are currently trying to commission and deploy a physical machine with MAAS Terraform provider.

We have a code block to create a `maas_machine` resource which will commission and detect the hardware inventory. 
Now when it comes to storage provisioning, it seems using a different value for the `name` attribute in the `maas_block_device` resource, other than the one detected during commissioning would create a new disk and leave stale disk `sda`.  

Here `sda` is the original disk name, and we are trying to name it `osdisk0` instead. As per the screenshot below I am ended up having additional block_device. 

Below is the `maas_block_device` resource definition. 

```
resource "maas_block_device" "osdisk1" {
    depends_on = [ terraform_data.clear_storage_layout ]
    machine = maas_machine.node04.id
    name = "osdisk0"
    id_path = local.osdisk[0].id_path
    size_gigabytes = 111
    partitions {
        size_gigabytes = 1
        fs_type        = "fat32"
        mount_point    = "/boot/efi"
        bootable = true
    }

    partitions {
        size_gigabytes = 100
    }
}
```

Image

On the contrary, I can rename the `maas_network_interface_physical` for network resources. I can set name to `enx8` instead of the original name as shown in the screenshot. I think mac address is used as the identifier here. 

```
resource "maas_network_interface_physical" "enx8" {
    machine = maas_machine.node04.id
    mac_address = "8c:ae:4c:fd:a4:43"
    name = "enx8"
}
```

Image

Expected Outcome - Ideally, I would expect to use id_path or serial number + model name as unique identifiers. So that I can update the device name, similar to network interfaces.

skyecas:
## Description of changes

Introduce the searching by id_path as well as name, to attempt to find a disk definition that is preexisting before falling back to creating a new disk, to prevent disk renames from creating resources.

fixes: #396 

## Checklist

### DIFF
--- a/maas/resource_maas_block_device.go
+++ b/maas/resource_maas_block_device.go
@@ -199,14 +199,11 @@ func resourceBlockDeviceCreate(ctx context.Context, d *schema.ResourceData, meta
 		return diag.FromErr(err)
 	}
 
-	// 1. SMART SEARCH: Check if the disk already exists based on our hierarchy
-	// (id_path > model&serial > name)
 	blockDevice, err := findBlockDeviceFromSchema(client, machine.SystemID, d)
 	if err != nil {
 		return diag.FromErr(err)
 	}
 
-	// 2. CREATE (If not found): If our smart search returned nil, it truly doesn't exist.
 	if blockDevice == nil {
 		blockDevice, err = client.BlockDevices.Create(machine.SystemID, getBlockDeviceParams(d))
 		if err != nil {
@@ -214,10 +211,8 @@ func resourceBlockDeviceCreate(ctx context.Context, d *schema.ResourceData, meta
 		}
 	}
 
-	// 3. SAVE STATE: Lock in the immutable MAAS database ID
 	d.SetId(fmt.Sprintf("%v", blockDevice.ID))
 
-	// 4. SYNC CONFIG: Pass it to the UpdateContext to apply any name/tag changes
 	return resourceBlockDeviceUpdate(ctx, d, meta)
 }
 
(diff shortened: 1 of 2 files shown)


maas/resource_maas_block_device.go
```diff
@@ -336,6 +341,43 @@ func getBlockDeviceParams(d *schema.ResourceData) *entity.BlockDeviceParams {
...
+		if model != "" && serial != "" && b.Model == model && b.Serial == serial {
+			return &blockDevices[i], nil
+		}
+
+		if name != "" && b.Name == name {
+			nameMatch = &blockDevices[i]
+		}
+	}
+
+	if nameMatch != nil {
```
skatsaounis: why de we have to do this comparing to other if clauses where we return immediately?

maas/resource_maas_block_device_test.go
```diff
@@ -75,6 +79,22 @@ resource "maas_block_device" "test" {
 `, machine)
 }
 
+func testAccMAASBlockDeviceRename(machine, name, idPath string) string {
```
skatsaounis: Nit: A better name would be something that mentions the creation by id_path. The `Rename` suffix is misleading since the rename is happening one level above, when this function is called with same idPath but different name

maas/resource_maas_block_device.go
```diff
@@ -199,20 +199,25 @@ func resourceBlockDeviceCreate(ctx context.Context, d *schema.ResourceData, meta
 		return diag.FromErr(err)
 	}
 
-	blockDevice, err := findBlockDevice(client, machine.SystemID, d.Get("name").(string))
+	// 1. SMART SEARCH: Check if the disk already exists based on our hierarchy
```
skatsaounis: Nit: I would delete all comments and add comments instead on top of function name, to document what the `findBlockDevice` does. Especially steps 2,3,4 are very obvious and can be omitted.

skatsaounis (changes requested):

maas/resource_maas_block_device.go
```diff
@@ -336,6 +336,42 @@ func getBlockDeviceParams(d *schema.ResourceData) *entity.BlockDeviceParams {
...
+		if model != "" && serial != "" && b.Model == model && b.Serial == serial {
+			return &blockDevices[i], nil
+		}
+
+		if name != "" && b.Name == name {
+			nameMatch = &blockDevices[i]
+		}
+	}
+
+	return nameMatch, err
```
skatsaounis: nit:

```suggestion
	return nameMatch, nil
```

skatsaounis (dismissed):
Approved with a nit

skatsaounis (approved):

PR #475, opened 2026-07-21
feat: add skip api checks

tobiasdemendonca:
## Description of changes

For Terragrunt stacks, we need the ability to create a plan without an available MAAS. As we make API calls during the provider configuration, `terragrunt stack run plan` currently fails. This PR adds a flag to disable API calls during the configure phase. 

This comes with some downsides like only being able to discover at apply time, not plan time, if a plan is valid for a particular MAAS version, but the benefits of having a plan without requiring a running MAAS outweigh this.

`skip_api_checks` is meant to be the flag that a terragrunt unit can use to turn off all config time api calls. If in future we add more api calls, then those changes can be turned off under this generic flag so units don't have to change as the provider increases in version. 

Related but does not complete: https://github.com/canonical/maas-terraform-modules/issues/113

## Checklist


tobiasdemendonca:
The current implementation means that the version is _never_ read into the config when this flag is enabled, so if someone were to use this flag with it always on, version checks for resources won't occur.

docs/index.md
```diff
@@ -46,6 +46,7 @@ The provider API block contains the necessary credentials to allow Terraform to
 - `api_url` (String) The MAAS API URL (eg: http://127.0.0.1:5240/MAAS). If not provided, it will be read from the MAAS_API_URL environment variable.
 - `api_version` (String) The MAAS API version (default 2.0)
 - `installation_method` (String) The MAAS installation method. Valid options: `snap`, and `deb`.
+- `skip_api_checks` (Boolean) Skip all checks that make API calls to MAAS during the provider configuration phase. This allows the provider to create a plan without a running MAAS present, such as when using Terragrunt stacks. This will potentially allow invalid plans, so use with caution. This currently only skips getting the current MAAS version, so resource version compatibility can only be determined at apply time directly from MAAS.
```
skatsaounis: I would like to avoid referring to Terragrunt in the provider registry docs. Is this an option?

skatsaounis (changes requested):
LGTM, except from the reference to Terragrunt.

skatsaounis (approved):
LGTM

PR #481, opened 2026-08-18
fix: allow dns_servers to be a computed field

tobiasdemendonca:
## Description of changes

In https://github.com/canonical/terraform-provider-maas/pull/446 , the behaviour introduced introduces a breaking change. `dns_servers` no longer acted as a `Computed` field, despite being labelled as one. This change was never released so there is no issue associated with it. This PR fixes this, by allowing a user to remove all dns_servers from their subnet by explicitly setting `dns_servers=[]` as before, but when omitted from the plan, allows a user to remove `dns_servers` from their resource and it to become a computed field.

This allows this to be a non-breaking change, as users upgrading won't have a non-empty plan. 

Please see the comments below for additional explanation and QA.

NA

## Checklist


skatsaounis (approved):
LGTM

